Issue #12, opened 2016-10-28
Settable property indication


lordkokkei:
I like that property/settable indicates that a property can be set. However, it may not be obvious for the front-end app to figure out which values or ranges that the property can be set to, so I suggest the following:

homie/686f6d6965/switch/$properties → on:settable[true,false]
indicates that a switch "on" can be set to "true" or "false"

homie/686f6d6965/temperature/$properties → temperature:settable[60-100]
indicates that the temperature (for example, a sous vide machine) can be set between 60 to 100 (unit defined elsewhere)

This would make settable properties more clear and obvious.

marvinroger:
I do agree!

Your suggestion handles the following:

* Enums
* Integer ranges

What about strings (which can be anything, not only enums), floats, more specific types like colors (`255,255,255`) etc.?

I would rather define data types. Each data types could have optional parameters (like min, max...). Something like:

homie/686f6d6965/switch/$properties → on:settable[enum:true,false]
homie/686f6d6965/temperature/$properties → temperature:settable[integer:60,100]

What do you think?

marvinroger:
I would even replace the `[]` with `()`, to differentiate ranges and data types:

homie/686f6d6965/ledstrip/$properties → led[1-3]:settable(enum:true,false)

lordkokkei:
Good thinking with data types. I completely agree with what you suggested.

Strings should be relatively simple to handle. For example, setting a color using predefined names can be done like this:
homie/686f6d6965/ledstrip/$properties → led[1-3]:settable(string)
homie/686f6d6965/ledstrip/led_1/set ← red

Multivariable settings are slightly tougher, but I think using a separator should be able to handle it, like so:
homie/686f6d6965/ledstrip/$properties → led[1-3]:settable(integer: 0,255; 0, 255; 0,255)

This would indicate that there are three settable integers for this property, each within the range [0, 255]. The color can be set like this:
homie/686f6d6965/ledstrip/led_1/set ← 0, 127, 255

ThomDietrich:
Hey @lordkokkei, I had the same problem as you did and wrote a few lines over in the [openHAB community](https://community.openhab.org/t/framework-homie-for-esp8266/8674/255?u=thomdietrich) because I overlooked your discussion here.

I do agree, that announcing the **unit, the datatype and the range** of one property should be standardized in the homie convention. I would however choose another approach than defining a mixed string, which is both hard to parse and easy to mistake and on top of that not using what MQTT has to offer.

[...]

PR #27, opened 2017-07-31
Added Attributes to node properties

kwave:
In order to releave @marvinroger a little bit, I've created a PR which contains the changes discussed in #12 . Let's proceed with the discussion directly in the PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ homie/686f6d6965/$fw/version → 1.0.0
 
 ### Node properties
 
-* `homie` / **`device ID`** / **`node ID`** / **`property`**: `node ID` is the ID of the node, which must be unique on a per-device basis, and which adhere to the [ID format](#id-format). `property` is the property of the node that is getting updated, which must be unique on a per-node basis, and which adhere to the [ID format](#id-format).
+* `homie` / **`device ID`** / **`node ID`** / **`property`** / **`attribute`**: `node ID` is the ID of the node, which must be unique on a per-device basis, and which adhere to the [ID format](#id-format). `property` is the property of the node that is getting updated, which must be unique on a per-node basis, and which adhere to the [ID format](#id-format).
 
 Properties starting with a `$` are special properties. It must be one of the following:
 
@@ -176,22 +176,99 @@ Properties starting with a `$` are special properties. It must be one of the fol
   <tr>
     <td>$properties</td>
     <td>Device → Controller</td>
-    <td>Properties the node exposes, with format <code>id</code> separated by a <code>,</code> if there are multiple nodes. For ranges, define the range after the <code>id</code>, within <code>[]</code> and separated by a <code>-</code>. For settable properties, add <code>:settable</code> to the <code>id</code></td>
+    <td>Properties the node exposes, with format <code>id</code> separated by a <code>,</code> if there are multiple nodes. For ranges, define the range after the <code>id</code>, within <code>[]</code> and separated by a <code>-</code>.</td>
     <td>Yes</td>
     <td>Yes</td>
   </tr>
 </table>
+
+#### Attributes
+An attribute describes various aspects of a property and therefore makes it auto-discoverable by any other device. A property can be described using the following attributes:
+<table>
+    <tr>
+        <th>Attribute</th>
+        <th>Direction</th>
+        <th>Description</th>
+        <th>Valid values</th>
+        <th>Retained</th>
+        <th>Required</th>
+    </tr>
+    <tr>
+        <td>$settable</td>
+        <td>Device → Controller</td>
+        <td>Specifies whether the property is settable (<code>true</code>) or readonly (<code>false</code>)</td>
+        <td><code>true</code>,<code>false</code></td>
+        <td>Yes</td>
+        <td>Yes</td>
+    </tr>
+    <tr>
+        <td>$unit</td>
+        <td>Device → Controller</td>
+        <td>A string containing the unit of this property. You are not limited to the recommended values, although they are the only well known ones that will have to be recognized by any Homie consumer.</td>
+        <td>
+            Recommended: <br>
+            <code>°C</code> Degree Celsius<br>
+            <code>°F</code> Degree Fahrenheit<br>
+            <code>°</code> Degree<br>
+            <code>L</code> Liter<br>
+            <code>gal</code> Galon<br>
+            <code>V</code> Volts<br>
+            <code>W</code> Watt<br>
+            <code>A</code> Ampere<br>
+            <code>%</code> Percent<br>
+            <code>m</code> Meter<br>
+            <code>ft</code> Feet<br>
+            <code>#</code> Count or Amount
+        </td>
+        <td>Yes</td>
+        <td>Yes</td>
+    </tr>
+    <tr>
+       <td>$datatype</td>
+       <td>Device → Controller</td>
+       <td>Describes the format of data.</td>
+       <td><code>integer</code>, <code>float</code>, <code>boolean</code>, <code>string</code>, <code>enum</code> </td>
+       <td>Yes</td>
+       <td>Yes</td>
+    </tr>
+    <tr>
+       <td>$range</td>
+       <td>Device → Controller</td>
+       <td>
+        Describes what are valid values for this property.
+       </td>
+       <td>
+             <ul>
+                <li><code>from:to</code> Describes a range of values e.g. <code>10:15</code>. <br>Valid for datatypes <code>integer</code>, <code>float</code> </li>
+                <li><code>value,value,value</code> for enumerating all valid values. Escape <code>,</code> by using <code>,,</code>. e.g. <code>A,B,C</code> or <code>ON,OFF,PAUSE</code>. <br>Valid for datatypes <code>enum</code> </li>
+                <li><code>/regex/</code> to provide a regex that can be used to match the value. e.g. <code>/[A-Z][0-9]+/g</code>. <br>Valid for datatype <code>string</code></li>
+            </ul>
+       </td>
+       <td>Yes</td>
+       <td>Yes</td>
+    </tr>
+    
+    
+</table>
+
 
 For example, our `686f6d6965` above would send:
 
 ```
 homie/686f6d6965/temperature/$type → temperature
 homie/686f6d6965/temperature/$properties → degrees,unit
-homie/686f6d6965/temperature/unit → c
+homie/686f6d6965/temperature/degrees/$settable → false
+homie/686f6d6965/temperature/degrees/$unit → C
+homie/686f6d6965/temperature/degrees/$datatype → float
+homie/686f6d6965/temperature/degrees/$range → -20.0:60
 homie/686f6d6965/temperature/degrees → 12.07
 
 homie/686f6d6965/humidity/$type → humidity
 homie/686f6d6965/humidity/$properties → percentage
+homie/686f6d6965/humidity/percentage/$settable → false
+homie/686f6d6965/humidity/percentage/$unit → %
+homie/686f6d6965/humidity/percentage/$datatype → integer
+homie/686f6d6965/humidity/percentage/$range → 0:100
 homie/686f6d6965/humidity/percentage → 79
 ```
 
@@ -199,9 +276,24 @@ A LED strip would look like this. Note that the topic for a range properties is 
 
 ```
 homie/ledstrip-device/ledstrip/$type → ledstrip
-homie/ledstrip-device/ledstrip/$properties → led[1-3]:settable
+homie/ledstrip-device/ledstrip/$properties → led[1-3]
+
+homie/ledstrip-device/ledstrip/led_1/$settable → true
+homie/ledstrip-device/ledstrip/led_1/$unit → LED Status
+homie/ledstrip-device/ledstrip/led_1/$datatype → enum
+homie/ledstrip-device/ledstrip/led_1/$range → on,off
 homie/ledstrip-device/ledstrip/led_1 → on
+
+homie/ledstrip-device/ledstrip/led_2/$settable → true
+homie/ledstrip-device/ledstrip/led_2/$unit → LED Status
+homie/ledstrip-device/ledstrip/led_2/$datatype → enum
+homie/ledstrip-device/ledstrip/led_2/$range → on,off
 homie/ledstrip-device/ledstrip/led_2 → off
+
+homie/ledstrip-device/ledstrip/led_3/$settable → true
+homie/ledstrip-device/ledstrip/led_3/$unit → LED Status
+homie/ledstrip-device/ledstrip/led_3/$datatype → enum
+homie/ledstrip-device/ledstrip/led_3/$range → on,off
 homie/ledstrip-device/ledstrip/led_3 → on
 ```
 
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ You can find an implementation of the Homie convention:
 
 ## Background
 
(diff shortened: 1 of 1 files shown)


ThomDietrich:
Hey @kwave,
thanks for taking up this task!
Big coincidence, just this weekend I've developed an MQTT speaking daemon that uses "Homie Convention v2.1.0-alpha" 😃 and I was able to gain some head-on experience.

Not a complete review but two early addition:
* I believe you missed to add `homie/device-id/$nodes` to the list of "Device Properties"
* I realized that  "`$name` Friendly name" should also be avilable for properties
* The required flag should not be Yes as often as it is currently. The settable property should for example have a default false nature instead of requiring the topic. Other topics are similarly not needed with a proper default.

README.md
```diff
@@ -176,32 +176,124 @@ Properties starting with a `$` are special properties. It must be one of the fol
...
+            <code>°</code> Degree<br>
+            <code>L</code> Liter<br>
+            <code>gal</code> Galon<br>
+            <code>V</code> Volts<br>
+            <code>W</code> Watt<br>
+            <code>A</code> Ampere<br>
+            <code>%</code> Percent<br>
+            <code>m</code> Meter<br>
+            <code>ft</code> Feet<br>
+            <code>#</code> Count or Amount
```
euphi: I miss a recommendation for (air) pressure. I propose Hectopascal (Unit `hPa`), because it is a (derived) SI-unit, other than Bar or mBar.  You may add "PSI" as imperial unit.

README.md
```diff
@@ -176,32 +176,124 @@ Properties starting with a `$` are special properties. It must be one of the fol
...
+homie/ledstrip-device/ledstrip/led_2/$datatype → enum
+homie/ledstrip-device/ledstrip/led_2/$range → on,off
 homie/ledstrip-device/ledstrip/led_2 → off
+
+homie/ledstrip-device/ledstrip/led_3/$settable → true
+homie/ledstrip-device/ledstrip/led_3/$unit → LED Status
+homie/ledstrip-device/ledstrip/led_3/$datatype → enum
+homie/ledstrip-device/ledstrip/led_3/$range → on,off
 homie/ledstrip-device/ledstrip/led_3 → on
 ```
```
euphi: The following (unchanged) section is inconsistent with the LED example.  I like the new definition with "On" or "Off" as state, so I propose to update the following section to make use of it. (No more "On"-state that is set to true or false).

README.md
```diff
@@ -176,32 +176,124 @@ Properties starting with a `$` are special properties. It must be one of the fol
...
+        </td>
+        <td>Yes</td>
+        <td>Yes</td>
+    </tr>
+    <tr>
+       <td>$datatype</td>
+       <td>Device → Controller</td>
+       <td>Describes the format of data.</td>
+       <td><code>integer</code>, <code>float</code>, <code>boolean</code>, <code>string</code>, <code>enum</code> </td>
+       <td>Yes</td>
```
euphi: Format of boolean should be defined. (What is allowed, is it case sensitive?) "true" and "false" or also "TRUE", "FALSE", 0, 1?

README.md
kwave: I've added Pascal `Pa`(all other units are also listed in their in basic unit) and `psi`.

README.md
kwave: Good point. I've added this to the datatype.

README.md
kwave: Why is it inconsistent? It describes how a client can set a property of the homie device, and how the device gives feedback about whether the value has been processed or not. Please clarify.

kwave:
Fixed most of the stuff mentioned above.

> - I realized that "$name Friendly name" should also be available for Nodes
I've added this for nodes and also properties.
 

> - The required flag should not be Yes as often as it is currently. The settable property should for example have a default false nature instead of requiring the topic. Other topics are similarly not needed with a proper default.
I thought the same when writing down the changes. But there is one caveat that came to my mind: MQTT is message based. If you subscribe to ´property/#` you will receive a single message for each retained topic in **any** order. So how does the client know whether an attribute is not present (--> Use the default!) or if the retained message is about to be delivered later (--> Use the attribute value instead of the default). Any Ideas?

> - You've chosen a different regex format. Any reason why?
Your suggestion was lacking the possibility to add flags to the pattern. Thats all ;-)


> - In the README the convention structure is confusing. Maybe you could rename the section "Device properties" to "Device" and describe the deviceID topic as well as all homie/deviceID/$keyword topics under this section. The same goes for "Nodes" and for "Properties". As I mentioned in the original issue I also don't like the confusion around Property and properties. My suggestion would be that "Property" should only be used for the hierarchy level under Nodes (e.g. temperature) and the special .../$keyword topics should be addressed with "parameter", "meta-topic" or similar
I've changed the naming. Every subtopic starting with `$` is called "attribute" whereas the level below a node is still called "property"

README.md
```diff
@@ -35,14 +35,17 @@ An ID MAY contain only lowercase letters from `a` to `z`, numbers from `0` to `9
 To efficiently parse messages, Homie defines a few rules related to topic names. The base topic you will see in the following convention will be `homie/`. You can however choose whatever base topic you want.
 
 * `homie` / **`device ID`**: this is the base topic of a device. Each device must have an unique device ID which adhere to the [ID format](#id-format).
```
ThomDietrich: "a unique"

README.md
euphi: The next section in README.MD is unchanged and thus somehow inconsistent:

```
Homie is state-based. You don't tell your smartlight to turn on, but you tell it to put it's on state to true. This especially fits well with MQTT, because of retained message.

For example, a kitchen-light device exposing a light node would subscribe to homie/kitchen-light/light/on/set and it would receive:

homie/kitchen-light/light/on/set ← true
```

According to your propasal it is better to use

`homie/kitchen-light/light/state/set ← on`

README.md
```diff
@@ -18,7 +18,7 @@ You can find an implementation of the Homie convention:
...
+An instance of a physical piece of hardware (an Arduino, an ESP8266...) is called a **device**. A device has **attributes**, like the current local IP, the Wi-Fi signal, etc. A device can expose multiple **nodes**. For example, a weather device might expose a `temperature` node and an `humidity` node. A node can have multiple **properties**. The `temperature` node might for example expose a `degrees` property containing the actual temperature, and an `unit` property. Node properties can be **ranges**. For example, if you have a LED strip, you can have a node property `led` ranging from `1` to `10`, to control LEDs independently. Node properties can be **settable**. For example, you don't want your `degrees` property to be settable in case of a temperature sensor: this depends on the environment and it would not make sense to change it. However, you will want the `degrees` property to be settable in case of a thermostat.
```
ThomDietrich: The example is not a good fit anymore. Unit is now it's dedicated attribute after all.
To illustrate the usage well we need more levels.

I'm not an expert on weather stations but how about something along this example:

* homie/
  * weatherstation/
    * base/
      * temperature
      * humidity
    * windsensor/
       * direction
       * speed
 * coffeemachine/
    * ...

...Plus all the additional attributes

README.md
```diff
@@ -35,14 +35,17 @@ An ID MAY contain only lowercase letters from `a` to `z`, numbers from `0` to `9
 To efficiently parse messages, Homie defines a few rules related to topic names. The base topic you will see in the following convention will be `homie/`. You can however choose whatever base topic you want.
 
 * `homie` / **`device ID`**: this is the base topic of a device. Each device must have an unique device ID which adhere to the [ID format](#id-format).
+* Attributes are topics that are prefixed with a `$`. These sub-topics add meta-data to Devices, Nodes and Properties describing their parent topic.
 
-### Device properties
 
-* `homie` / **`device ID`** / `$` **`device property`**: a topic starting with a `$` after the base topic of a device represents a device property. A device property MUST be one of these:
+### Device attributes
```
ThomDietrich: My suggestion was to move the sentence from above regarding the device below this headline and rename the headline to "Device". For the new reader it is pretty confusing that there is a headline e.g. "Node ...." and it immediately addresses attributes while the Node is not yet discussed. I'd suggest four big headlines "Homie Base", "Devices", "Nodes", "Properties".

ThomDietrich:
One general remark that might go beyond your PR: As you can see git management of long paragraphs is rather unfitting. Because of that it's often recommended to break down paragraphs into "one sentence per line" code. The markdown result will be the same but the git management of the document is better.

README.md
ThomDietrich: agreed. Alternatively `homie/kitchen-light/light/power/set ← on`

README.md
```diff
@@ -4,7 +4,7 @@
 
 **Please note this v2 branch is a work-in-progress. It might change before the final release.**
 
-Version: **2.0.0**.
+Version: **2.1.0**.
```
marvinroger: As written in the above line, the v2 is not yet final, so we don't even need to bump the version. 👍

README.md
```diff
@@ -139,6 +142,13 @@ To efficiently parse messages, Homie defines a few rules related to topic names.
     <td>Yes or No, depending of your implementation</td>
     <td>No</td>
   </tr>
+  <tr>
+    <td>$nodes</td>
+    <td>Device → Controller</td>
+    <td>Nodes the device exposes, with format `id` separated by a `,` if there are multiple nodes. For ranges, define the range after the `id`, within `[]` and separated by a `-`.</td>
```
marvinroger: Can you replace the `` ` `` with `<code>` tags? The `` ` `` doesn't work in HTML arrays.

README.md
marvinroger: Good point.

README.md
marvinroger: Agreed.

README.md
marvinroger: Agreed too. @kwave feel free to choose between `state` and `power`. 😉

README.md
```diff
@@ -174,34 +183,147 @@ Properties starting with a `$` are special properties. It must be one of the fol
...
 homie/686f6d6965/humidity/$type → humidity
 homie/686f6d6965/humidity/$properties → percentage
+homie/686f6d6965/humidity/percentage/$settable → false
+homie/686f6d6965/humidity/percentage/$unit → %
+homie/686f6d6965/humidity/percentage/$datatype → integer
+homie/686f6d6965/humidity/percentage/$range → 0:100
 homie/686f6d6965/humidity/percentage → 79
 ```
 
 A LED strip would look like this. Note that the topic for a range properties is the name of the property followed by a `_` and the index getting updated:
```
euphi: The term _range_ is now used in two different meanings in the convention.

1. To allow "arrays" of properties.
2. To define allowed value range

**--> possible Solution:** change `$range` to `$valuerange`
An alternative would be to change to range term (as used in the "old" convention) to something else, e.g. _vector_ .

README.md
```diff
@@ -174,34 +183,147 @@ Properties starting with a `$` are special properties. It must be one of the fol
...
-homie/ledstrip-device/ledstrip/$properties → led[1-3]:settable
+homie/ledstrip-device/ledstrip/$properties → led[1-3]
+
+homie/ledstrip-device/ledstrip/led_1/$settable → true
+homie/ledstrip-device/ledstrip/led_1/$unit →
+homie/ledstrip-device/ledstrip/led_1/$name → Red LEDs
+homie/ledstrip-device/ledstrip/led_1/$datatype → enum
+homie/ledstrip-device/ledstrip/led_1/$range → on,off
 homie/ledstrip-device/ledstrip/led_1 → on
+
```
euphi: Personally, I don't think it is a good idea to set the attribute of each property range-index individually. $settable, $unit, $datatype and $range (or $valuerange) should be the same for the complete range of the property and thus be defined on property level.
Ranges (or _vectors_) should simplify things and not make them more complicated. If you need individual attributes, use individual properties.

There may be a use for naming each ranged property:
 - I/O modules of industrial automation systems usually use formal names based on a simple numbering scheme (comparable to homie's _range_ (or _vector_)), but allow to set an optional symbolic name.

README.md
kwave: Ok done. Replaced "on" with power in all topics.

README.md
kwave: I've modified the example to be more precise. Basically added a new property for the battery level to the example.

README.md
kwave: Fixed.

README.md
kwave: I would suggest to version the smaller increments anyways. As the final version is still a v2, what about 2.0.1 for the changes discussed here?

README.md
kwave: Strange, in my Atom-Markdown preview this works as expected. I've changed it.

README.md
kwave: Agreed, this might get somehow confusing. I've changed the `$range` to `$format`, as "range" is the wrong term anyways. e.g. a regex is definitely not a "range", same for enums.

README.md
ThomDietrich: Thanks but it's still not how I imagine it. Feel free to ask if my suggestion wasn't clear. Also I'm not saying that my suggestion is the only and best solution, so let me know if you disagree with anything ;)

README.md
ThomDietrich: Be aware, that you missed out on the "base" and "windsensor" nodes. With the text as it is right now the meaning of nodes and properties doesn't get clear.

ThomDietrich:
Hey guys, great progress so far!
I did a little brainstorming with a colleague today and we found room for two improvements I'd like to discuss with you:

1. **Property class/category:** A property is a distinct reading or characteristic value produced by a node of a device. The other side of the network, the discovering entity, needs to know about the nature of this value (e.g. to show a thermometer in a GUI). We are already providing attributes to characterize its unit, range, pretty name and "set'ability" but not a clear type.
  A not-settable property with the unit "°C" does at best *imply* that the property is a temperature reading of a thermometer, not e.g. a thermostat or a nominal temperature. The discovering entity should not have to resort to guessing and interpreting the other attributes.
  I propose a **counted number of property classes** to characterize the nature of a property. The `$class` or `$category` attribute is required but can be "Unspecified"/"Other"/"" for special non-trivial properties.

2. **Conversion Attribute:** When working with embedded systems or low-energy micro controllers, we can not expect the device to be capable to convert a raw sensor reading into a final human-comprehendible value. An example would be a temperature sensor reading between [0..255] which would map onto a degree Celcius value with decimal places.
  I propose the introduction of a `$conversion` property attribute. The published String value is either "x"/"" for no conversion or the right hand of an `f(x)` formula.

Both points are of course open for discussion. What do you think?

marvinroger:
@ThomDietrich I see a problem with *property* classes. Imagine you have an `heater` node, with two properties: `mode` (`off,comfort,night`), and `temperature`. How do you assing a "class" to `mode`, for example? The most logical thing to do would be to assing classes to *nodes*, not *properties*. Therefore we would have `heater`, `light`, `weatherstation`... But this would introduce another problem: for the UI to be able to show a thermometer with the temperature embedded in the picture, we would have to define a standard set of properties for each class. Which defeat the flexibility of the homie convention, right?

For the conversion attribute, I am not that sure. It's rather complicated, and I don't see in practice what kind of value can't be computed directly on-device. I mean, even a 3$ chip is able to read a temperature in `°C`, `°F`.

README.md
marvinroger: I would say no, but as the v2 is already widely used, let's keep it this way.

README.md
marvinroger: `$format` seems perfect. @euphi ?

marvinroger:
@kwave 

> Your suggestion was lacking the possibility to add flags to the pattern. Thats all ;-)

A simple regex pattern is enough, there's no need to use any flags in such a case, right? (global, unicode, multiline...). The only exception being the *ignore flag*, but for ease of implementation by parser and simplicity, I think it's better to do `[a-zA-Z]` rather than `/[a-z]/i`. Objections?

euphi:
I didn't had the time yet to look at @ThomDietrich 's proposal, but I suggest to move this discussion to an own issue, because it is out of scope of the original intention of this PR, so this PR can be merged soon.

ThomDietrich:
Agreed. Let's merge the PR soon and incorporate further additions later.
There are many details I'd like to add or incorporate and individual PRs for these, based on the state of this PR now would be easier to manage...

PR #34, opened 2017-08-07
Made clear that all values are converted to string

euphi:


ThomDietrich:
Good summary! Not sure if "Value" is the right headline and if this is really a part of the Topology rundown. How about a new level 2 headline below "ID Format"?

marvinroger:
Indeed, the @ThomDietrich approach seems better. @euphi do you agree?

euphi:
Yes, I agree :+1:

marvinroger:
Fixed by https://github.com/marvinroger/homie/commit/11cd9ff521070cc2c99bc481c9a0536c6f0fa811

ThomDietrich:
@marvinroger the sentence is a bit too vague for a convention for my taste.
Numbers should be defined with their exact representation (delimiter, exponential components etc., we could go with normal rational numbers with the dot delimiter, e.g. 123.456 or 89).
Boolean should accept multiple variants (true, TRUE, 1, ON).

euphi:
Yes, I agree. The sentence is a good start, but needs to be refined.

For numbers, I propose to use the default LC_NUMERIC behavior.

I looked for some standard and found  [ISO/IEC 30112](http://www.open-std.org/JTC1/SC35/WG5/docs/30112d10.pdf), Page 68f..

However, it seems that the decimal_point is a comma in the standard, which is not what most programming language have set as default, so it would introduce confusion.

So, I propose to use a dot as decimal_point and no grouping or seperators.

exponential components should be out of scope of the convention. If necessary, numbers that need them, can be transported as implementation-specific string.

PR #35, opened 2017-08-07
$unit should not be mandatory

euphi:
For example, ENUMs and Strings are usually unit-less.


euphi:
Alternatively, there should be a clarification that $unit is mandatory and the payload MUST be empty if the information is unit-less. (Thats how it is done in the example right now, but "_empty_"  is not a valid value for "$unit" yet.

ThomDietrich:
Other attributes should be optional as well. I was about to create a PR for that later today.

marvinroger:
Ah, can you make a PR against `redesign`?

PR #36, opened 2017-08-07
Improve example presentation with strings and java syntax

ThomDietrich:


marvinroger:
> The java without quotes is the prettiest IMHO. But the quotes might make the users think that "quoted strings" are of type string and that the rest are actually boolean and numbers, except it's not (everything is a string)

Don't you agree?

ThomDietrich:
https://github.com/marvinroger/homie/pull/31#discussion_r131542178

Not sure what you mean. The java highlighting alone will color everything a bit random and make mistakes at dots. Also as discussed with @euphi it should be clear if integers or boolean variables are communicated as bytes or as translated strings. If the one or the other is better is still worth discussing. However at the moment everything is converted to a string, so that needs to be shown clearly. Hence the second option: 

java:

```java
homie/686f6d6965/$homie → 2.1.0
homie/686f6d6965/$name → Weather station
homie/686f6d6965/$localip → 192.168.0.10
homie/686f6d6965/$stats/uptime → 120
homie/686f6d6965/$online → true
```

java with "quotes":

```java
homie/686f6d6965/$homie → "2.1.0"
homie/686f6d6965/$name → "Weather station"
homie/686f6d6965/$localip → "192.168.0.10"
homie/686f6d6965/$stats/uptime → 120
homie/686f6d6965/$online → true
```

marvinroger:
My bad, I misunderstood. I thought you only wanted to quote strings (and not numbers even if its still a string), which would have been confusing. Anyway, thanks! :+1:

Issue #24, opened 2017-06-14
$online - unexpected connection lost


timpur:
Wondering if its better to use the last will to post information about unexpected connection lost

Currently disconnecting correctly and unexpected are the same = false at $online

Wondering if its better to use a number, 1 = online, 0 = offline and -1 = unexpected connection lost ??

Tell me what you think.
Tim.

marvinroger:
Good point. What do you guys @kwave @ThomDietrich @euphi think about this?

I'd prefer an explicit text instead of code numbers. `$online` would not be a binary state anymore, so maybe we should change the attribute to `$state` or `$status` (don't know what's the most correct in english). The values could therefore be `online`, `offline` (clean) or `lost` (unexpected disconnection).

euphi:
`$state` or `$status` is good (I'm unsure which one is better, both are ok).

For me, `offline` is too generic for a clean disconnect, so maybe `disconnected` is better.
There may also be a state `sleep` for deepsleep. (However, this is a little bit problematic, because the state will stay in `sleep` if the device cannot reconnect for other reasons.

Also, you could send a state after MQTT connection:
 
 - `reconnect` if only MQTT connection and/or WiFi was lost
 - `reset` if device was resetted
 - `poweron` if there was a power failure
 - `wakeup` after deep sleep (On ESP8266 this seems to be possible from API using the struct `resetInfo` but it seems this is somehow buggy?)

Just after this, Homie shall send the `online` state. This allows to see a reset/reboot reason by subscribing to `$state`.

timpur:
I like these ideas. Maybe have both?
`State` is simply `on` or `off` ?
`Status` is more complex, `Connected` , `disconnected`, `lost`, `sleep`, etc. 
??
This way we have best of both worlds, do not need to change the standard, but also adding more information about the devices current status?

In dress to @euphi 
> `reconnect` if only MQTT connection and/or WiFi was lost
`reset` if device was resetted
`poweron` if there was a power failure
`wakeup` after deep sleep (On ESP8266 this seems to be possible from API using the struct resetInfo but it seems this is somehow buggy?)

[...]

Issue #17, opened 2017-02-28
Extra device properties


bertmelis:
The convention now states a limited list of device properties.
I would however want to add a few more: a device could have a battery level, cpu load, free memory, temperature,...

For not to break the convention, I now create a "device-node" with the node properties as above, which I find not entirely correct.

(It is working however, so if the community -and off course you Marvin- doesn't agree with me, I'll stick to my current solution.)

euphi:
There should be a standard for the device node :-)

So, in Homie Convention there is an optional node (e.g. "$deviceinfo") with a standardized set of properties.

The devicenode could be implemented as extra library for various platforms, e.g. Homie-esp8266-DeviceInfo, so everyone is free to decide whether to use it.

bertmelis:
I'm not following.  Did I overlook something in the PR?

marvinroger:
I also don't get it 🤔

euphi:
@bertmelis the PR currently in discussions does not define more detailed device properties. In your proposal you ask for properties like CPU load, battery state, free mem etc.

Based on this I propose to define an optional "$deviceinfo" node, that is like a normal node, but with standardized properties, e.g.

`homie/myDevice/$deviceinfo/CPUtemperature --> 32`

Where CPUtemperature is a standardized property, so there is no need to send all the  unit info etc.

ThomDietrich:
May I ask what we would need this for?

The Homie convention as an MQTT convention is here to support auto discovery of devices and applications on an MQTT broker through a standardized way of announcing and publishing data and actions of that entity. Data like the CPU temperature is exactly that, data. It can be published by defining a node (e.g. CPU) and a property (e.g. temperature). No need for special attendance. Am I correct?

I did btw also not get your initial request @bertmelis, did you talk about something like standardized nodes, e.g. similarly to http://www.eclipse.org/smarthome/documentation/development/bindings/thing-definition.html#channel-categories ??

bertmelis:
Well, the Homie convention now gives a number of stats ($stats/uptime, signal, interval) which are also merely data. Signal is even optional. They are all device and configuration specific and do not relate to the thing being automated but to the device doing the automation.

The paragraph says device attributes MUST the one of the listed.

So It was thinking about making the list less strict and give some space to the $stats/# branch.

marvinroger:
[...]

PR #40, opened 2017-08-08
Moved statistics to seperate section and made some clarifications, when to send device attributes.

euphi:


euphi:
Just a proposal, feel free to improve (or decline..).

ThomDietrich:
I like the idea but then you might want to move the `$stats/interval` to the last row and add a note like "Interval in seconds at which the `$stats/` subtopics are refreshed (see below)"

README.md
```diff
@@ -150,16 +150,83 @@ A device attribute MUST be one of these:
...
+homie/686f6d6965/$stats/uptime → "120"
+homie/686f6d6965/$stats/interval → "60"
+homie/686f6d6965/$fw/name → "weatherstation-firmware"
+homie/686f6d6965/$fw/version → "1.0.0"
+homie/686f6d6965/$nodes → "outdoor-probe,windsensor"
+homie/686f6d6965/$implementation → "esp8266"
+homie/686f6d6965/$online → "true"
+```
+
+### Device statistics
```
ThomDietrich: should be h4?

README.md
```diff
@@ -150,16 +150,83 @@ A device attribute MUST be one of these:
...
+* `homie` / `device ID` / `$stats`/ **`$device-statistic-attribute`**:
+The `$stats/` hierarchy allows to send device attributes that change over time. They MUST be sent every `$stats/interval` seconds.
```
ThomDietrich: "MUST be sent" is a onesided view from the eyes of a IoT developer. We should also cover the Controller view.

README.md
```diff
@@ -101,7 +101,7 @@ Each device must have a unique device ID which adhere to the [ID Format](#id-for
 #### Device Attributes
 
 * `homie` / `device ID` / **`$device-attribute`**:
-A device attribute MUST be one of these:
+The set of device attributes MUST be sent immediately after connection to MQTT broker is established and MUST be one of these:
```
ThomDietrich: "MUST be sent" is a onesided view from the eyes of a IoT developer. We should also cover the Controller view.

ThomDietrich:
I've left two more comments. Looks good so far ;)

bertmelis:
I've been offline for a while (almost complete disconnect during holidays in "la douce France"). But this almost completely fits what I was thinking in #17 . I would just add a note to the table that it is exemplary and it is implementation-dependant. (as the "MUST be one of these" is left away).

Also, I would change <code>They MUST be sent every `$stats/interval` seconds.</code> to <code>All attributes MUST be sent every `$stats/interval` seconds.</code>

README.md
euphi: Yes, I already noticed that while writing - I wanted to avoid the passive sentence.
However, there are various other sections in the same wording, so I didn't want to change them all.
Furthermore, it is better to move the "trigger" to the beginning of the sentence.

So, better:

"When the MQTT connection to the broker is established or re-established, the device MUST sends its attributes to the broker immediately."

Regarding "immediately": I propose to create a new section for handling of the`$online` topic (or `$state` if it gets renamed), that clarifies the requirements for connection establishment (based on #24 ).

README.md
euphi: I don't think so. Device attributes are sent only once (per MQTT connection), while Device statistics are sent frequently - that's why I choosed to use h3.

README.md
ThomDietrich: it's a plan 👍

README.md
ThomDietrich: Ah I see... However look at this TOC:

grafik

The all containing headline is "Topology" and that makes total sense. In my eye statistics is part of device (topology'ly speaking) or it needs to move out of the Topology section, which would also be okay with me. We'd need to test that.

euphi: Ah, I thought "Device Attributes" was on H3. So, changing Device Statistics to h4 is correct :+1:

ThomDietrich:
Last comment: The example already includes interval and uptime but no others. Maybe move the example behind the stats section and add 1-2 of the new stats topics?

README.md
```diff
@@ -150,16 +150,82 @@ A device attribute MUST be one of these:
...
     <td>$stats/interval</td>
     <td>Device → Controller</td>
-    <td>Interval in seconds at which the <code>$stats/+</code> are refreshed</td>
+    <td>Interval in seconds at which the device refreshes its <code>$stats/+</code>: See next section for details about statistical attributes</td>
+    <td>Yes</td>
+    <td>Yes</td>
+  </tr>
+</table>
+
+For example, a device with an ID of `686f6d6965` with an `outdoor-probe` and a `windsensor` node would send:
```
ThomDietrich: with an... with an...

README.md
```diff
@@ -208,59 +274,15 @@ A device attribute MUST be one of these:
...
-    <td>$implementation/#</td>
-    <td>Controller → Device or Device → Controller</td>
-    <td>You can use any subtopics of <code>$implementation</code> for anything related to your specific Homie implementation.</td>
-    <td>Yes or No, depending of your implementation</td>
-    <td>No</td>
-  </tr>
 </table>
 
-For example, a device with an ID of `686f6d6965` with an `outdoor-probe` and a `windsensor` node would send:
+For example, a device with an ID of `686f6d6965` with `$stats/$interval` value "60" may send every 60seconds its current values:
```
ThomDietrich: 60seconds
may?
Suggestion: "is supposed to send it's current values every 60 seconds."

euphi: ok.

README.md
euphi: second "with an" changed to ".. that comprises of .."

euphi:
Maybe its a good idea to use a collaborative editor for further improvements.
I uploaded the current revision (of this PR) to https://hackmd.io/s/HJtf_lqDb . 
However, I think it would be good to merge first and than start to use hackmd.io for collaborative editing.

PR #47, opened 2017-09-25
Clean up MQTT restrictions

ThomDietrich:


README.md
```diff
@@ -29,28 +29,27 @@ The Homie convention is thereby a crucial aspect in the support of **automatic d
 
 ----
 
-Homie communicates through [MQTT](http://mqtt.org) and is hence based on the basic principles of MQTT topic publication and subscription.
-
-## ID Format
+## MQTT Restrictions
```
ThomDietrich: Open for other suggestions

README.md
```diff
@@ -29,28 +29,27 @@ The Homie convention is thereby a crucial aspect in the support of **automatic d
...
-
 The underscore (`_`) is used and reserved for Homie *node arrays*.
 
-## Payload
+### Payload
 
-Every MQTT message payload MUST be sent as string. If a value is of a numeric data type, it MUST be converted to string. Booleans MUST be converted to "true" or "false". All values MUST be encoded as UTF-8 strings. 
+Every MQTT message payload MUST be sent as string.
+If a value is of a numeric data type, it MUST be converted to string.
+Booleans MUST be converted to "true" or "false".
```
ThomDietrich: @marvinroger do you think this restriction is wise? I'd suggest to allow: "1", "true", "TRUE", "on", "ON"

euphi: I prefer to have only one allowed "string value" for each converted value, so "true" or "false" for Booleans. This helps parsing a lot.

Why did you remove the UTF-8 requirement? If you send a string over Homie, I think it is good to have a strict requirement for the encoding.

ThomDietrich: I did not remove UTF8. It's below this comment ;)

I'd also prefer to have just one possible option. Sadly the world is not perfect and Homie is a convention to connect many systems. The way how a boolean value is communicated might differ between implementations. A standard openHAB MQTT configuration would for example send "ON" as a command.

We should differentiate the direction of the communication:

- A Homie node will publish its states (e.g. $settable) as either `true` or `false`
- A controller sending a property command to a Homie node may send either of the above variations.

davidgraeff: And homie describes itself as being lightweight in the very first sentences. The controllers have to adapt. The burden should not be on the small-memory µC.

PR #48, opened 2017-09-25
Make properties optional with default

ThomDietrich:


ThomDietrich:
In a future version of the convention we should replace the table by free text to have more options to add notes etc.

README.md
```diff
@@ -426,7 +426,7 @@ A property attribute MUST be one of these:
             <code>#</code> Count or Amount
         </td>
         <td>Yes</td>
-        <td>Yes.<br /> If the property is unit-less, e.g. a discrete state, set this to <code>none</code></td>
+        <td>No ("")</td>
```
marvinroger: A MQTT message with a zero-length content will delete the associated retained message, which is why we chose `none`

ThomDietrich: Understood.
However, the default is not what is published onto the MQTT topic, the default is the value that will be used by a controller IF the topic is NOT published.

ThomDietrich:
@marvinroger just checked this PR once again. I believe it is okay as it is.

As I said before, we should think about replacing the table in a later version.

README.md
marvinroger: Oh, got you. Did not notice the `(default)`

marvinroger:
@ThomDietrich don't you think @oeysteinhansen is right when he says `$name` device attribute should be optional, etc.? These are not required.

My initial idea was that every Homie messages should be required in order for a controller to know when the announce is complete. Now that there are default values, this assumption is false. So we can probably slim the required messages to the absolute minimum. @ThomDietrich you're probably the one with the most hindsight about controller implementation, so what do you think?

ThomDietrich:
Hey Marvin,
this was a misunderstanding on my side. The PR as it is is only addressing Property attributes. @oeysteinhansen could you please create a PR following my example for the device attributes? Thanks!

ThomDietrich:
Regarding discovery and announcement finish: I'm also not an expert in the area, just invested some thought into it 😃 My goal is to have Homie as the first auto-discoverable MQTT based convention in openHAB as a binding.
I believe that a complete announcement would only make sense if we were able to define a context-free language. As soon as context is needed (i.e. the Homie convention plus a specific version) we can rely on that context (e.g. default values). @oeysteinhansen argued correctly, this will reduce complexity in the client node and hence will easy the pain of client implementation - this *might* help with broader adoption.
Regarding the end of announcement: The addition of the `state` property will help with that step. We might have to look out for user reports: I fear problems with propagation delays or queue shuffling. It might be in the responsibility of the controller to react on announcement messages after the `ready` state is reached.

In this context you may see, why #51 is important to me...

marvinroger:
@ThomDietrich definitely, an implementation guideline is needed for broader adoption.

marvinroger:
@oeysteinhansen let's create another PR for device attributes. 😉

PR #60, opened 2018-01-17
Update README.md

nerdfirefighter:
Fixed error in temperature/$properties example.
Also added $name attribute to the nodes and properties.

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,18 +286,22 @@ For example, our `686f6d6965` above would send:
 
 ```
 homie/686f6d6965/temperature/$type → temperature
-homie/686f6d6965/temperature/$properties → degrees,unit
+homie/686f6d6965/temperature/$name → Bedroom Temperature Node
+homie/686f6d6965/temperature/$properties → degrees
 homie/686f6d6965/temperature/degrees/$settable → false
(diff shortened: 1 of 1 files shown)


timpur:
@nerdfirefighter, really you should be making these changes in the redesign branch because we will be merging that will be the final version.

Thinking we might revert 2.0 to 2.0 status and 2.1 / redesign will be going to a new branch called v3.0

or might merge it all into a dev and master convention, and make releases of the spec, which i prefer my self.

nerdfirefighter:
@timpur i don't disagree, however i also think that documentation related to older/not in development versions/release should be accurate and that is my reason for this pull request. 

As for repo layout, i think a similar layout to the esp-8266 repo would be good. given this is more of a documentation repo rather than a code repo it might be simpler to have a structure of:
dev-vX - where X is the major version. It would be current to the latest release
vX - where X is the major version. any minor releases could be done using the release markers like in the esp8266
i'm not a git expert so i am not sure what the 'correct' or 'best practice' git layout is for this, just that a branch called redesign is not necessarily the easiest to identify when that redesign was happening (between which versions)

timpur:
I'll accept this and merge in to the redesign branch. Thanks :)

(When on a computer next)

PR #64, opened 2018-02-10
Add MicroPython implementation

rroemhild:


timpur:
why 'WIP' ?

rroemhild:
It's not that stable we want it to be. But it can be used right now. I think for the issues we see, we can add a know issue topic to our README.md.

timpur:
Okay sounds good

rroemhild:
Known Issue section and the reference to Homie v2.1.0 were added. Free to merge.

Issue #46, opened 2017-09-25
Repository branching and tagging, versioning


ThomDietrich:
Hello @marvinroger,

Somewhere in comments we already discussed that branching in this repository should be addressed. I want to restart that discussion. Here's my first suggestion, which is open for discussion:

1. **Introduce a strict versioning system** - Homie is a convention to be followed by other development projects. It is hence important to add unambiguous version numbers for every convention revision. A developer needs to be able to state "The *app* conforms with the Homie convention v2.1.3". I suggest to implement semantic versioning: http://semver.org (should be a perfect fit)

2. **Work in the `main` branch only** - Currently there is an outdated master branch, an unclear v2.0 branch and a redesign branch. That is neither intuitive nor visible to a reader. As the convention is steadily evolving, we should present the latest revision to the interested developer/user. (With the option to switch to other versions, see 5)

3. **Use tags to mark versions** - A homie version is a fixed state of the document, hence tags are perfectly fine for the job. (Other [branching models](http://nvie.com/posts/a-successful-git-branching-model/) suggest branches in order to allow e.g. hotfixes)

4. **Version every commit** - Almost every commit will change details of the Homie convention, it is therefore needed to increment the version number as the MAJOR.MINOR.PATCH system of semver suggests. (Exceptions are of course commits with language changes only)

4. **GitHub releases for stable tags** - GitHub offers the [releases tab](https://github.com/marvinroger/homie/releases) to publish versions and add a release note. That should be used to announce and document the stable and noteworthy releases (e.g. v1.5.0, v2.0.0, v2.1.0)

5. **Link the "Tags" tab in the first part of the convention** - For readers to quickly find all available homie versions, the [tags tab](https://github.com/marvinroger/homie/tags) should be highlighted in the first part of the convention document.

Everyone, please add your comments, criticism and feedback!
Best! Thomas 🍻

marvinroger:
Hi @ThomDietrich 
Thanks for that! Actually, this was already planned (we do exactly all of that on https://github.com/marvinroger/homie-esp8266). I wanted to "clean everything" as soon as we all agreed about the v2.1.0, which will be the last release that breaks semver (the v2.0.0 is not really backward compatible, but I'd say this is fine as the v2.0.0 was never released officially).

The only thing I did not think about was the "Tags" tab. Great idea! 👍

marvinroger:
https://github.com/marvinroger/homie/commit/e0e3bad3c659c12146819d53b2612098acedc176 ✨

[...]

PR #69, opened 2018-03-21
Merge redesign into master

ThomDietrich:
According to #46 

@timpur @euphi  whatever the next steps might be, I believe this is the first thing we need to do. Please add a 👍 if you agree. I'll do the merge manually to have a cleaner resulting graph.

(You can ignore the long list of commits and existing comments, this will be rather clean after the merge)


timpur:
Let's wait till we come to a decision as a group.

PR #73, opened 2018-04-01
Add my ESP32 IDF implementation

lukecyca:


ThomDietrich:
Nice one!

May I ask if your implementation covers the whole convention (as of v2.0.1) to your best knowledge?
Are you planning to adapt to v3.x.x soon?

lukecyca:
I think it _technically_ conforms to the full v2.0.1 specification, in that sends every topic that is required, and doesn't send anything illegal.

The main limitation is that it isn't very convenient to add new stats and nodes (the application needs to do it manually). I plan to work on that in the coming days though, so I think it is safe to say it conforms.

I will migrate to v3 eventually, but I will wait for the esp8266 framework and the homie-ota projects to go first, since I personally have both esp8266 and esp32 and want to control them both in the same way.

PR #88, opened 2018-04-27
Remove regex requirement

davidgraeff:
* A good regex parser is not lightweight and most likely needs lots of heap memory for the state machine and backreferences and whatnot. Do not make it mandatory to support it.
* Be more precise when the property format attribute is required

### DIFF
--- a/README.md
+++ b/README.md
@@ -495,10 +495,6 @@ A property attribute MUST be one of these:
              <br>Valid for datatypes <code>enum</code>
            </li>
            <li>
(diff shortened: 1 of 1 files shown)


marvinroger:
I have to agree, it's not that useful anyway. Never had to use it in practice.

It does not break semver, so it does not harm. @timpur @euphi @ThomDietrich LGTM?

ThomDietrich:
Fine with me 👍

timpur:
All for !!!

marvinroger:
Thanks guys!

PR #89, opened 2018-04-27
Use "true" instead of "on"

davidgraeff:
I guess part of this convention is not only how devices/attributes/nodes/properties present themselves, but also to a certain degree structure and define messages. Be consistent with the example where a boolean value is sent. Use "true" instead of "on".

### DIFF
--- a/README.md
+++ b/README.md
@@ -530,7 +530,7 @@ This especially fits well with MQTT, because of retained message.
 For example, a `kitchen-light` device exposing a `light` node would subscribe to `homie/kitchen-light/light/power/set` and it would receive:
 
 ```java
-homie/kitchen-light/light/power/set ← "on"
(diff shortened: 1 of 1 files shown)


timpur:
Not sure about this

davidgraeff:
The example is not very specific, but I expect the "power" property to be defined like this:
```
homie/mydevice/lamp/power/$name → "Lamp 1"
homie/mydevice/lamp/power/$settable → "true"
homie/mydevice/lamp/power/$datatype → "boolean"
homie/mydevice/lamp/power → "true"/"false"
```

ThomDietrich:
I have to agree. "on" doesn't differ from "true" semantically.
Both could be not ideal. Power being "true" is just as unnatural as a door-sensor being "on".

Also see: https://github.com/homieiot/convention/pull/47#discussion_r140890829

PR #94, opened 2018-05-02
More realistic array example

davidgraeff:
More realistic array example. You are using arrays if you have tons of subjects, not just two

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ Nodes are independent or logically separable parts of a device.
 For example, a car might expose a `wheels` node, an `engine` node and a `lights` node.
 
 Nodes can be **arrays**.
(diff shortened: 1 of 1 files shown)


ThomDietrich:
Is that all? The `lights` example is used more than once and you'd need to adapt them all.
I like your intention but an ledstrip doesn't match @marvinroger 's example of an RC car.

Maybe easier: You could leave the original text and add the led strip example below.

timpur:
If your going to make the effort to make a change, may as well enhance the examples with it ect... ?
Sorry David.

davidgraeff:
The RC car should have a ledstrip then :)

You can name the elements of the given array ("backlight", "frontlight"), it feels more like an anti-example of an array.

ThomDietrich:
How about

grafik

davidgraeff:
The Array spec gets an overhaul soon anyway, I guess. Closing this PR

Issue #101, opened 2018-05-05
Regarding simplification and additional node wishes


davidgraeff:
Tags: Organisation, Convention/Spec document layout

It looks quite immature for the outsider if there are major version bumps of the spec like every some months. A modular system might help:

Thing of a Homie Core 1.0, Homie Device Statistics 1.0, Homie System Information 1.0 and so on. If one of those parts change (version bump), the rest can still be understood by a controller. The current situation is: `/home/device/$localip` moves to somewhere else or gets a different semantic and because of that and a 4.0 version bump of the entire spec, a controller is not able to do discover device nodes anymore? Doesn't sound right.

### The Homie Core 1.0 convention
My suggestion is: Throw everything out of the core spec that is value related. Only device, node and property discovery should be part of the the core. No statistics, no ip, no mac. No suggested/recommended "$type"s.

### Homie Device statistics 1.0 spec
I suggest to introduce a new kind of node (prefixed with "_"). Those are defined by a related specification.
Add a "_stats" node with all those optional properties ("cpuload", "battery" etc). The advantage is: No special case for the controller anymore. If battery is a percentage, then set `homie/device/_stats/battery/$format`=`0:100` and the controller knows how to render a property, even if additional properties are specified later on.

The controller might need to know the version. Add `homie/device/_stats/$version`=`1.0`.

###  Homie System Information 1.0 spec
Basically all those ip, mac, soc, app version, os version, service ports and so on information. Stuff that is party necessary for updates or direct communication.

The controller might to know the version. Add `homie/device/_system/$version`=`1.0`.

### Homie Node Types 1.0
Define some node and property types for controllers to render nodes and properties correctly. Something like:
* **light** `homie/device/mk12a/$type`=`light`: Needs "power". Optional: "color", "dim"
* **switch** `homie/device/gang1/$type`=`switch`: Needs "power".

### Split the convention into multiple files
The core document should link to those other files in prominent places but should not contain anything related expect examples.

### Split up the repository
Every of those mentioned specs should get an own repository for issue discussing and preparing spec documents which are then merged into the main repository on full member agreement. Almost no issue here is about the core spec anymore. It is mostly about additional nodes, additional types and so on.

fermuch:
I'm with you, @davidgraeff, on having a minimal spec and then branching out to "how to implement complex stuff with homie", but more as tutorials instead of complementary specs. “How can I implement a home alert system with Homie?” “How can I implement a smart refrigerator with Homie?”

[...]

PR #104, opened 2018-05-07
Homie Core convention

davidgraeff:
A concrete implementation of the Homie Core convention idea, expressed in #101 for discussion.

Signed-off-by: David Graeff <david.graeff@web.de>

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,68 +158,12 @@ When the MQTT connection to the broker is established or re-established, the dev
     <td>Yes</td>
   </tr>
   <tr>
-    <td>$localip</td>
-    <td>Device → Controller</td>
-    <td>IP of the device on the local network</td>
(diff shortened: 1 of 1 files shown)


timpur:
lets slow down a bit and think things through a bit more, we need to make sure this is the right choice and i dont think @ThomDietrich has said much, and tbh i haven't had time i think about this (even though my thoughts isn't the decider). This is a big change and requires proper discussion tbh.

ThomDietrich:
I'm supposed to be in the final weeks of my PhD thesis and broke my arm one week ago. Please do not expect much from me right now. @davidgraeff I'm sorry I didn't engage so far. Let's push #101 forward!

timpur:
@ThomDietrich, oh my, yes please focus on your PhD.

PR #114, opened 2018-10-28
Remove parts that belong on a web page instead

davidgraeff:
Signed-off-by: David Gräff <david.graeff@web.de>


marvinroger:
@davidgraeff can you open a PR against `develop` instead?

Issue #111, opened 2018-10-28
Confusion on device/$state message order


boc-tothefuture:
The documentation states (Emphasis is mine):

> init: this is the state the device is in when it is connected to the MQTT broker, but has not yet sent all Homie messages and is not yet ready to operate. **This is the first message** that must that must be sent.

The example right above it shows:

> homie/super-car/$homie → "2.1.0"
homie/super-car/$name → "Super car"
homie/super-car/$localip → "192.168.0.10"
homie/super-car/$mac → "DE:AD:BE:EF:FE:ED"
homie/super-car/$fw/name → "weatherstation-firmware"
homie/super-car/$fw/version → "1.0.0"
homie/super-car/$nodes → "wheels,engine,lights[]"
homie/super-car/$implementation → "esp8266"
homie/super-car/$stats/interval → "60"
homie/super-car/$state → "ready"

Based on how I interpret the documentation, the example should be:

```
homie/super-car/$state → "init"
homie/super-car/$homie → "2.1.0"
homie/super-car/$name → "Super car"
homie/super-car/$localip → "192.168.0.10"
homie/super-car/$mac → "DE:AD:BE:EF:FE:ED"
homie/super-car/$fw/name → "weatherstation-firmware"
homie/super-car/$fw/version → "1.0.0"
homie/super-car/$nodes → "wheels,engine,lights[]"
homie/super-car/$implementation → "esp8266"
homie/super-car/$stats/interval → "60"
homie/super-car/$state → "ready"
```
But then it seems odd to send the state before the version of the protocol?  Is the documentation off or am I interpreting this incorrectly?

lorenwest:
Great eagle-eye. If there is to be value in `$state` having both `init` and `ready` values, I think it should be documented to be the 2nd post-connect message, with `$homie` version being the first.

If `$homie` version is output first, that gives us the ability to understand the meaning of `$state`, and even have the opportunity to change it in future versions. 

Not that we would, but we could.

davidgraeff:
All those topics are retained. The broker is not required to publish them in order. My openHAB implementation is order agnostic, I think that is the best approach.

lorenwest:
Then the suggestion would be to remove references to any ordering from the spec? I think that's a good idea.

boc-tothefuture:
@davidgraeff - good point on the topics not having to be sent by the broker in a defined order.

[...]

PR #115, opened 2018-10-29
Fixed wording on init state.

boc-tothefuture:
Updated init state wording to indicate the state was optional in the lifecycle and provided an example of when it may be used.

PR based on discussion of #111


marvinroger:
Hi @boc-tothefuture 

We're in the process of moving the "development" version to the `develop` branch. Would you mind opening a PR against `develop` instead?

boc-tothefuture:
Closed.  Opened #116 against develop branch.

PR #118, opened 2018-10-29
Remove parts that belong on a web page instead. License.

davidgraeff:
* Move convention text to a new file "convention.md"
* Modify readme to contain the motivation for the convention
  and a link to the webpage.
* Change license to "Creative Commons Attribution 4.0 International Public License"
  which works for specifications, unlike the MIT.

Signed-off-by: David Gräff <david.graeff@web.de>


convention.md
```diff
@@ -1,53 +1,6 @@
...
-
-<p align="center">Version: <b><!--VERSION-->3.x.x<!--VERSION--></b> • <a href="https://github.com/homieiot/convention/tags">Browse Versions…</a></p>
-
-**![Rolling Release](https://cdn2.iconfinder.com/data/icons/thesquid-ink-40-free-flat-icon-pack/64/traffic-light-20.png) The Homie Convention follows the rolling release model and is improved and released in short development cycles. Check the *releases* tab to learn about previous release versions.**
-
-Implementations of the Homie convention can be found on [this page](implementations.md).
-
-----
-
-## Table of Contents
```
timpur, line 14: Personally i still think the table of contents is handy. Usefully for when you want to find a cretin bit in the convention.

timpur (commented):
Looks good. :)

davidgraeff:
I suggest to change the license to the CC-4. The creative commons license was made for texts, images and media in general. The MIT license talks about "software" and might not apply here.

This is probably something all maintainers should agree on.

convention.md
```diff
@@ -1,53 +1,6 @@
...
-
-MQTT supports easy and unrestricted message-based communication.
-However, MQTT doesn't define the structure and content of these messages and their relation.
-An IoT device publishes data and provides interaction possibilities but a controlling entity will need to be specifically configured to be able to interface with the device.
-
-The Homie convention defines a **standardized way** of how IoT devices and services announce themselves and their data on the communication channel.
-The Homie convention is thereby a crucial aspect in the support of **automatic discovery, configuration and usage** of devices and services over the MQTT protocol.
-
-----
+Version: **x.x.x**
```
marvinroger: I'd keep the title and the version centered. Also, I think the banner makes the document a bit prettier.

convention.md
marvinroger: The ToC is kind of hard to maintain. It's available on the website, and it can be automatically generated on the `master` version, anyway. 😉

marvinroger:
Regarding the license, I'm not very aware, to be honest! But it seems to make more sense.

convention.md
timpur: Even better :)

timpur:
I'm not valid to comment about licence tbh, I approve of the rest :).

convention.md
davidgraeff: The banner is still there in the readme, just not in the specification file itself.

The version is just a one liner, non centered like in other ietf or ieee specifications.
Personally I would like to add an auto generated date (date placeholder) even. The website mentions the release date already within the version selection box.

marvinroger: Fine. Can you add a `Date: <!--DATE-->AUTO GENERATED<!--DATE-->`, and replace `Version: **x.x.x**` by `Version: **<!--VERSION-->3.x.x<!--VERSION-->**`? Thanks :)

Issue #121, opened 2018-11-01
Specification is unclear on empty string for property values


boc-tothefuture:
From reading the specification it is not clear to me if the convention supports empty/blank strings for property values. 

Using the example:
`homie/super-car/engine/temperature → "21.5"`
Does the convention support a later message of:
`homie/super-car/engine/temperature → ""`
 

This should be clarified, because if it is allowed, the controller implementations need to handle blank/empty strings.

In most cases it would work for Strings.  I don't know what numbers would do, and I am not sure what enums would do either.  It would essentially be "unsetting" that value.

I am coming across it in my kotlin device implementation as I consider what to do if the user passes null value.

It seems complicated to allow this - but right now, unless I missed it, the convention doesn't address this issue one way or the other.

davidgraeff:
An empty value except for strings is undefined behaviour. Properties can only be unset by setting the $properties attribute of the parent node.

boc-tothefuture:
Should we state in the convention that an empty string (besides string type) is disallowed?

davidgraeff:
You cannot disallow it unfortunately. Empty is the same as if the topic value is not published at all. And because of timing and order issues, it can happen that the property type is already published (let's say integer) and the value is not published yet. This is just undefined behaviour and a controller implementation must handle the situation gracefully.

boc-tothefuture:
@davidgraeff -
 Can you expand on that a bit.. I am afraid I am not following how empty is different than never published? If the value is not published yet, you wouldn't receive anything on the controller side?  If the value was published an an empty string is then published you would receive the empty string.

davidgraeff:
@boc-tothefuture What do you do in your controller implementation if everything but the actual property value has been received? Do exactly the same for an empty received property.

boc-tothefuture:
@davidgraeff -
 It probably depends on the actual controller implementation.  There could be some that do nothing (don't expose the received property) until they receive the first value for the property.  Doing that would be within the spec, AFAIK.

What would be the counterpoint to disallowing, per spec, the sending of empty strings for non-string values?

My opinion is that we should attempt to make the convention as clear as possible on what is allowed to be transmitted.  Items that lead to undefined behavior should be minimized to enable as broad of compatibility between controllers and devices as possible.

davidgraeff:
An integer property is allowed to publish integers only, a boolean one  true, false and so on. I don't really understand the problem. If you publish an empty string it is undefined behaviour, I don't think we need to explicitly say this, do we?

[...]

Issue #51, opened 2017-11-06
Implementation guideline for client/controller developer


ThomDietrich:
Derived from https://github.com/marvinroger/homie/issues/28#issuecomment-320690407

Besides all the technical definitions of the Homie convention, we should give simple checklist-like instructions for developers on how to comply with the Homie convention. There are many (technical and logical) details that need to be handled, which are not easily described in the convention itself.

I believe the guideline will further clarify the benefits of the Homie convention and ease the adaption for unacquainted developers.

@marvinroger Please label with "help wanted"

marvinroger:
Do you think we need to write this guideline before releasing the v3?

ThomDietrich:
No. Would be useful but it doesn't influence the convention. Can be v3.0.1 ;)

timpur:
@marvinroger and @ThomDietrich When do you think you'll release `v3.0` ? Should i start implementing the new spec or wait (`homie-esp8266`)?

timpur:
https://github.com/homieiot/convention/issues/93#issuecomment-385961536
> It is absolutely not about the semantic of an input or output. It is just about controllers knowing of all available inputs and ouputs and their data types (int, bool, string, etc).

Just to calcify @davidgraeff, this would have nothing to do with the spec, but more as help on how to use the spec if a dev wants it. It will help future people like you and me know where to start to discover all the homie devices :P as the title says: a `guideline `

ingoogni:
I'm getting utterly confused by this opening and closing and moving of discussion. Can't we have a decent mailinglist or even Gitter (meh)  for discussing things. This is worse than USENET 30y ago.

@davidgraeff thanks for "It is just about controllers knowing of all available inputs and ouputs and their data types (int, bool, string, etc" that clarifies somewhat.
What does homie does not do yet? What is not possible yet? Does it need programming? Is it a matter of standardising certain types of topics? Is there an MQTT auto-discovery convention? Would that be useful?

ThomDietrich:
This is the way how issue ticket systems work for thousands to millions of projects on all kinds of platforms. They normally work quite well as long as people don't mix issues and discuss one single topic in one single issue.

davidgraeff:
@ingoogni The spec is definitely not written for someone that has never seen MQTT before or has never thought about M2M communication. It does its best to introduce to the topic though and I would say it does a good job. All your questions are actually answered explicitly or implicit.

[...]

PR #127, opened 2018-11-04
FIX: Numeric payload decimal separator is a dot

davidgraeff:


convention.md
```diff
@@ -18,10 +18,9 @@ The underscore (`_`) is used and reserved for Homie *node arrays*.
 
 ### Payload
 
-Every MQTT message payload MUST be sent as string.
-If a payload is of a numeric data type, it MUST be converted to string.
+Every MQTT message payload MUST be sent as UTF-8 encoded string.
+A numeric data type MUST contain a numeric value published as string with the decimal separator, if any, being a dot.
```
ThomDietrich: Reads as a bit confusing.

```suggestion
A numeric data type MUST be published as a string with a dot as the decimal separator, e.g. "123.456".
```

davidgraeff: The important point for @boc-tothefuture is that an integer payload can never be an empty string, that is somehow lost in your sentence. And the decimal separator is not a requirement for numerical values, only if there is one, it should be the dot.

ThomDietrich: The first should be implicit (because "" is not a valid numeric !?). If we want to emphasize this I'm sorry to tell you that I didn't understand this from your version.
Regarding the second: Imho a decimal delimiter is always only needed if decimal places are present. If you want to specifically state that maybe it would be best to divide the whole sentence in two.

davidgraeff: Actually I was happy with the original sentence (I just noticed that we don't specify the decimal separator), but it was found to be inaccurate in #121. @boc-tothefuture could you please present your argumentation here?

boc-tothefuture: It is my opinion that we should specify that empty values, besides for string type are not allowed. I could see an implementation thinking that sending an empty string for a property payload _clears_ the value.  I don't think that, but I could see how someone might.

For example, is an empty string OK for type enum?  The convention doesn't say that it is not, but I also think we can't actually create it because the way that we escape commas in the parsing of enum definitions.

For the comment that "" is not a valid numeric.  I agree, but some languages don't.  For example in ruby `"".to_i` happily returns 0.  If someone had only a ruby background, they might think an empty payload is acceptable to represent zero for a numeric.

This is beyond the scope of this item pull request, but I was going to put together a PR to more clearly define the acceptable values. 

As for this particular PR, I think defining the dot as the decimal separator is good.

ThomDietrich: Empty messages on retained topics delete the retained topic. This is most probably something we don't want in a homie client <--> controller communication, even though client and server should be able to handle this case.

The convention should disallow empty values - however not just for numbers. The change in this PR should clarify that each property can only be set within the range of its defined datatype. For numbers this is at least one digit, for enums this is exactly 1-of-n and for strings this is any combination of 0..* characters (incl. empty string).

In #51 it might be a good idea to emphasize this requirement.

As for this PR, @davidgraeff let me know if you would like to see a new suggestion from me.

davidgraeff: @ThomDietrich I definitely welcome another suggestion.

ThomDietrich:
```suggestion
A numeric data type MUST be converted to a literal string with the dot character as the decimal separator for the `float` data type.
An empty message is not valid as payload for numeric data types.
The value published as payload MUST be valid for the respective property/attribute type as per the convention definition below.
```

Not totally happy with it. Feel free to further adapt and merge.

davidgraeff: Surely better / more precise than the original one

PR #132, opened 2018-11-07
Add $extensions for extension specification.

Thalhammer:


convention.md
```diff
@@ -127,6 +127,15 @@ When the MQTT connection to the broker is established or re-established, the dev
     </td>
     <td>Yes</td>
   </tr>
+  <tr>
+    <td>$specs</td>
```
davidgraeff: Hm. It sounds very amateur to name this $specs. It is an abbreviation and not even homie calls itself "specification", it is a convention.
Better words might be "extensions", "expansions", "scopes" or so.

convention.md
```diff
@@ -127,6 +127,15 @@ When the MQTT connection to the broker is established or re-established, the dev
...
+    <td>Device → Controller</td>
+    <td>
+      A list of optional specifications the device supports, e.g. OTA, with format <code>id</code> separated by a <code>,</code> if there are multiple specs.
```
davidgraeff: The table gets really filled if we include long descriptive texts. I was going to change that at some point. 
I suggest a more brief text like:

List of one or more supported extensions, separated by comma.
MUST be an empty string otherwise.

convention.md
```diff
@@ -475,3 +484,14 @@ homie/$broadcast/alert ← "Intruder detected"
 ```
 
 Any other topic is not part of the Homie convention.
+
+### Optional Specifications
```
davidgraeff: Should this maybe be prefixed with "Appendix: " / "Addendum:"? It is basically a further explanation of something mentioned in the main text.

The description what an extension is might even be in the "Device" section below the attribute table.

convention.md
```diff
@@ -475,3 +484,14 @@ homie/$broadcast/alert ← "Intruder detected"
...
+
+In addition to homie-core the device may support multiple extensions which are defined in separate specifications.
+Every extension is identified by a unique id. In order to make sure those ids are unique, there are rules on how they are created. An id should only contain url safe characters.
+
+#### Official extension
```
davidgraeff: I don't like this differentiation. Currently we have a paragraph titled "Topic ID" where we explain the format of a topic. An analogue paragraph about "Extension ID" should just explain that we use the "reverse domain name" convention and that "org.homie" or "homie." is reserved.

Thalhammer:
Changed some parts to reflect suggestions by @davidgraeff

convention.md
```diff
@@ -475,3 +484,9 @@ homie/$broadcast/alert ← "Intruder detected"
 ```
 
 Any other topic is not part of the Homie convention.
+
+### Optional Extensions
+
+In addition to homie-core the device may support multiple extensions which are defined in separate conventions.
+Every extension is identified by a unique id, which is similar to a topic id but also allows dots.
```
davidgraeff: "which is similar to a topic id but also allows dots" is superfluous, because in the next sentence you clarify/narrow down the allowed id to a reverse domain name. Maybe combine those two sentences? We should link to an official document specifying *reverse domain name*. I only found wikipedia explaining it here: https://en.wikipedia.org/wiki/Reverse_domain_name_notation but no real RFC, except the one that specifies URLs (https://tools.ietf.org/html/rfc3986#section-2)

convention.md
```diff
@@ -475,3 +484,9 @@ homie/$broadcast/alert ← "Intruder detected"
...
+The id consists of the reverse domain name of a vendor owned domain and a freely choosen prefix. For example a company owning `example.org` that likes likes to extend homie a cool feature would choose `org.example.cool-feature`. If it is a official extension the prefix must be `homie`.
```
davidgraeff: You don't need to own a domain to choose a unique reverse domain name and you can't make this mandatory in my opinion, if homie itself doesn't own its own reverse domain name namespace.

> If it is a official extension the prefix must be `homie`.

That sounds more like an internal instruction. The wording in other specifications is more like "The namespace x.y is reserved."

convention.md
boc-tothefuture: Why an empty string instead of not existing? 
Is this to handle the case where a device had extensions and no longer does?

convention.md
boc-tothefuture: The sentence has a typo "likes" appears twice in a row.

boc-tothefuture:
If the controllers do not support the extension, is there a defined behavior in the convention?  Do they emit an error, do they not subscribe to any of the devices topics, or do they process only the topics they understand per the convention?

convention.md
davidgraeff: Yes

davidgraeff:
@boc-tothefuture If something is implicit, I'd like to not add it explicitly to the convention. Nobody will read it anymore if it is 100 pages, because we explicitly write everything down that was given implicitly already.
The word "extension" clearly means that a controller can ignore one, two or all of the referenced extensions.

convention.md
Thalhammer: @boc-tothefuture If you had a device which announced extensions and replaced it with one which one that does not, the controller would still think the extension is supported as the topic would not get replaced.

davidgraeff (approved):

timpur:
@Thalhammer @davidgraeff: what about versioning of these extensions ? it was mentioned in the issue, but not here ??

davidgraeff:
That is not exactly specified. I imagine that an extension requires the device to publish the version to a $extensionName attribute like it is done by $homie.

Thalhammer:
@timpur There was the idea to include a version number the extension list, however, we came to the conclusion that it's better to not define a hard version schema and instead let the extension decide how to publish its version. This should not be a problem since if a controller does not know the extension it has no use for the version and if it knows the extension it should also know where to find its version.

timpur:
Just thinking from a controller perspective, when it knows that it needs to load an extension from `$extensions` how does it know which version of the extension to load with out loading an version discovery phase first. ?? It would be nice for a controller to know the name and version of the extension up front so it know which extension module to use to parse the data ? (only if supporting many versions in a controller is possible) 

Like currently to do it youd need to steps, look at `$extensions` then `$extensionsVersion` this might add unnecessary complexity for the controller since this all happens async ... ?

Thalhammer:
@davidgraeff What do you think?
@timpur Personally I think this should be handled by the extension, and since it needs to be async anyway I don't think it adds a lot of additional complexity.

timpur:
Just means that you have to partially load the extensions controller fist to understand where the version info is ... ? think @davidgraeff , is best to visualise (imagine) if this would be an unnecessary complexity or not, since hes build a functional controller.

Thalhammer:
> since hes build a functional controller

So did I ;)
And since @davidgraeff was leading the idea to remove the version numbers in the first place, I think he's fine with the current state.

ThomDietrich:
I also can't think of a reason why the version number is needed for extension identification. As @Thalhammer said in https://github.com/homieiot/convention/pull/132#issuecomment-441807560: A controller which knows an extension should be able to identify and work with extension version. Just as always: If the controller can't identify a set of messages to form a supported extension/convention/whatever -> ignore it.

Well done guys

timpur:
Sorry @Thalhammer, didnt know that, which one ?? 

Okay i retire my case.

Thalhammer:
@timpur https://github.com/Thalhammer/homie-cpp

It's both a client and controller implementation. Wasn't updated in a will but will get a rewrite once 4.0 is finalized.

PR #133, opened 2018-11-08
[WIP] Additional specificity on payload definitions

boc-tothefuture:
This addresses, among other things, the conversation started in #121 

Please note, that for things like enums, I am not sure if I think we should treat whitespace should differentiate enums - but that is my interpretation of the spec as it stands today.

### DIFF
--- a/convention.md
+++ b/convention.md
@@ -18,47 +18,47 @@ The underscore (`_`) is used and reserved for Homie *node arrays*.
 
 ### Payload
 
- - Every MQTT message payload MUST be sent as a UTF-8 encoded string
- - The value published as payload MUST be valid for the respective property/attribute type as per the defined convention
+- Every MQTT message payload MUST be sent as a UTF-8 encoded string
+- The value published as payload MUST be valid for the respective property/attribute type as per the defined convention
  
 String
- - String types are limited to 268,435,456 characters 
- - An empty string ("") is a valid payload
+- String types are limited to 268,435,456 characters 
+- An empty string ("") is a valid payload
  
 Integer
- - Integer types are UTF-8 encoded representations of 64-bit signed whole numbers
- - Integers range from -9,223,372,036,854,775,808 (-2<sup>63</sup>) to 9,223,372,036,854,775,807 (2<sup>63</sup>-1)
- - The payload may only contain whole numbers and the negation character "-". No other characters including spaces (" ") are permitted 
- - A string with just a negation sign ("-") is not a valid payload
- - An empty string ("") is not a valid payload
+- Integer types are UTF-8 encoded representations of 64-bit signed whole numbers
+- Integers range from -9,223,372,036,854,775,808 (-2<sup>63</sup>) to 9,223,372,036,854,775,807 (2<sup>63</sup>-1)
+- The payload may only contain whole numbers and the negation character "-". No other characters including spaces (" ") are permitted 
+- A string with just a negation sign ("-") is not a valid payload
+- An empty string ("") is not a valid payload
  
 Float
 - Float types are UTF-8 encoded representations of 64-bit signed floating point numbers
- - Floats range from 2<sup>-1074</sup> to (2-2<sup>-52</sup>)*2<sup>1023</sup>
- - The payload may only contain whole numbers, the negation character "-", the exponent character "e" or "E" and the decimal separator ".", no other characters, including spaces (" ") are permitted 
- - The dot character (".") is the decimal separator (used if necessary) and may only have a single instance present in the payload
- - Representations of numeric concepts such as "NaN" (Not a Number) and "Infinity" are not a valid payload
- - A string with just a negation sign ("-") is not a valid payload
- - An empty string ("") is not a valid payload
+- Floats range from 2<sup>-1074</sup> to (2-2<sup>-52</sup>)&ast;2<sup>1023</sup>
+- The payload may only contain whole numbers, the negation character "-", the exponent character "e" or "E" and the decimal separator ".", no other characters, including spaces (" ") are permitted 
+- The dot character (".") is the decimal separator (used if necessary) and may only have a single instance present in the payload
+- Representations of numeric concepts such as "NaN" (Not a Number) and "Infinity" are not a valid payload
+- A string with just a negation sign ("-") is not a valid payload
+- An empty string ("") is not a valid payload
  
 Boolean
- - Booleans must be converted to "true" or "false"
- - Representation is case sensitive, e.g. "TRUE" or "FALSE" are not valid payloads.
- - An empty string ("") is not a valid payload
+- Booleans must be converted to "true" or "false"
+- Representation is case sensitive, e.g. "TRUE" or "FALSE" are not valid payloads.
+- An empty string ("") is not a valid payload
  
 Enum
- - Enum payloads must be one of the values specified in the format definition of the property
- - Enum payloads are case sensitive, e.g. "Car" will not match a format definition of "car"
- - Payloads should have leading and trailing whitespace removed
- - An empty string ("") is not a valid payload
+- Enum payloads must be one of the values specified in the format definition of the property
+- Enum payloads are case sensitive, e.g. "Car" will not match a format definition of "car"
+- Payloads should have leading and trailing whitespace removed
+- An empty string ("") is not a valid payload
  
 Color
- - Color payload validity varies depending on the property format definition of either "rgb" or "hsv"
- - Both payload types contain comma separated whole numbers of differing restricted ranges
- - The encoded string may only contain whole numbers and the comma character ",", no other characters are permitted, including spaces (" ")
- - Payloads for type "rgb" contains 3 comma separated values of numbers with a valid range between 0 and 255. e.g. 100,100,100
- - Payloads for type "hsv" contains 3 comma separated values of numbers. The first number has a range of 0 to 360, the second and third numbers have a range of 0 to 100.  e.g. 300,50,75
- - An empty string ("") is not a valid payload
+- Color payload validity varies depending on the property format definition of either "rgb" or "hsv"
+- Both payload types contain comma separated whole numbers of differing restricted ranges
+- The encoded string may only contain whole numbers and the comma character ",", no other characters are permitted, including spaces (" ")
+- Payloads for type "rgb" contains 3 comma separated values of numbers with a valid range between 0 and 255. e.g. 100,100,100
+- Payloads for type "hsv" contains 3 comma separated values of numbers. The first number has a range of 0 to 360, the second and third numbers have a range of 0 to 100.  e.g. 300,50,75
+- An empty string ("") is not a valid payload
  
 
 ### QoS and retained messages
--- a/convention.md
+++ b/convention.md
@@ -50,7 +50,7 @@ Boolean
 Enum
  - Enum payloads must be one of the values specified in the format definition of the property
  - Enum payloads are case sensitive, e.g. "Car" will not match a format definition of "car"
- - Enum matches are exact, e.g. A payload of "car ", will not match a format definition of "car". " car", "car" and "car " Represent three valid and distinct enum values
+ - Payloads should have leading and trailing whitespace removed
  - An empty string ("") is not a valid payload
  
 Color
--- a/convention.md
+++ b/convention.md
@@ -19,21 +19,21 @@ The underscore (`_`) is used and reserved for Homie *node arrays*.
 ### Payload
 
 - Every MQTT message payload MUST be sent as a UTF-8 encoded string
-- The value published as payload MUST be valid for the respective property/attribute type as per the defined convention
+- The value published as payload MUST be valid for the respective property/attribute type as per the list below
  
 String
 - String types are limited to 268,435,456 characters 
 - An empty string ("") is a valid payload
  
 Integer
-- Integer types are UTF-8 encoded representations of 64-bit signed whole numbers
+- Integer types are UTF-8 encoded string literal representations of 64-bit signed whole numbers
 - Integers range from -9,223,372,036,854,775,808 (-2<sup>63</sup>) to 9,223,372,036,854,775,807 (2<sup>63</sup>-1)
 - The payload may only contain whole numbers and the negation character "-". No other characters including spaces (" ") are permitted 
 - A string with just a negation sign ("-") is not a valid payload
 - An empty string ("") is not a valid payload
  
 Float
-- Float types are UTF-8 encoded representations of 64-bit signed floating point numbers
+- Float types are UTF-8 encoded string literal representations of 64-bit signed floating point numbers
 - Floats range from 2<sup>-1074</sup> to (2-2<sup>-52</sup>)&ast;2<sup>1023</sup>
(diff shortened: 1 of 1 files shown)


convention.md
```diff
@@ -18,11 +18,50 @@ The underscore (`_`) is used and reserved for Homie *node arrays*.
...
+ - Every MQTT message payload MUST be sent as a UTF-8 encoded string
+-  The value published as payload MUST be valid for the respective property/attribute type as per the defined convention
+ 
+ #### String
+ - String types are limited to 268,435,456 characters 
+ - An empty string ("") is a valid payload
+ 
+ #### Integer
+ - Integer types are UTF-8 encoded representations of 64-bit signed whole numbers
+ - Integers range from -9,223,372,036,854,775,808 (-2<sup>63</sup>) to 9,223,372,036,854,775,807 (2<sup>63</sup>-1)
```
timpur: This looks like a signed 128bit number ?? Should it actually be a 64 bit signed number not 128 ?

convention.md
```diff
@@ -18,11 +18,50 @@ The underscore (`_`) is used and reserved for Homie *node arrays*.
...
+ - The payload may only contain whole numbers and the negation character "-". No other characters including spaces (" ") are permitted 
+ - A string with just a negation sign ("-") is not a valid payload
+ - An empty string ("") is not a valid payload
+ 
+ #### Float
+ - Float types are UTF-8 encoded representations of 64-bit signed floating point numbers
+ - Floats range from 2<sup>-1074</sup> to (2-2<sup>-52</sup>)*2<sup>1023</sup>
```
timpur: I dont know much about floats but this doesnt look right? https://en.wikipedia.org/wiki/IEEE_754-1985, they have `±2.23×10−308 to ±1.80×10308`

convention.md
boc-tothefuture: I am basing it off the numbers here: 
https://docs.oracle.com/javase/7/docs/api/constant-values.html#java.lang.Long.MAX_VALUE

convention.md
boc-tothefuture: Good find.  I took my values from here:  https://docs.oracle.com/javase/7/docs/api/java/lang/Double.html

convention.md
```diff
@@ -18,11 +18,50 @@ The underscore (`_`) is used and reserved for Homie *node arrays*.
...
+ #### Enum
+ - Enum payloads must be one of the values specified in the format definition of the property
+ - Enum payloads are case sensitive, e.g. "Car" will not match a format definition of "car"
+ - Enum matches are exact, e.g. A payload of "car ", will not match a format definition of "car". " car", "car" and "car " Represent three valid and distinct enum values
+ - An empty string ("") is not a valid payload
+ 
+ #### Color
+ - Color payload validity varies depending on the property format definition of either "rgb" or "hsv"
+ - Both payload types contain comma separated whole numbers of differing restricted ranges
+ - The encoded string may only contain whole numbers and the comma character ",", no other characters are permitted, including spaces (" ")
```
davidgraeff: The Homie 3 openHAB controller publishes values with whitespace.

convention.md
```diff
@@ -18,11 +18,50 @@ The underscore (`_`) is used and reserved for Homie *node arrays*.
 
 ### Payload
 
-Every MQTT message payload MUST be sent as UTF-8 encoded string.
-A numeric data type MUST be converted to a literal string with the dot character as the decimal separator for the `float` data type.
-An empty message is not valid as payload for numeric data types.
-The value published as payload MUST be valid for the respective property/attribute type as per the convention definition below.
-Booleans MUST be converted to "true" or "false".
+#### General Specification
```
davidgraeff: Please remove one level of headings each, and put everything under "General" directly under the payload topic.

bertmelis:
Watch out for empty strings. An empty payload will just delete the retained topic.

convention.md
boc-tothefuture: Should that be supported by the convention? If so, is it up to controllers to trim whitespace before attempting to parse payloads?

boc-tothefuture:
> Watch out for empty strings. An empty payload will just delete the retained topic.

An empty string seems like a valid value and represents a valid state.  For example, if I have a music system that has 4 lines of display and each line is a property.  There are some scenarios where line 4 may or may not have metadata.  If line 4 previously had metadata and no longer does, I think the appropriate action is to transmit an empty string.

davidgraeff:
By implication that means a controller must assume an empty string for all non existing topics.

convention.md
boc-tothefuture: done

convention.md
timpur: Dont know what i was thinking, your right that is a 64bit number lol. Think in arduino land a long is 32bit, and a long long is 64bit. We're okay with max 64bit number ?

boc-tothefuture: Not sure.  All of my definitions are up for discussion.  I created them based on my assumptions while reading the spec. I *think* a 64 bit number is ok.  Anything bigger than that gets complicated and may reduce targetable platforms.

convention.md
```diff
@@ -18,11 +18,49 @@ The underscore (`_`) is used and reserved for Homie *node arrays*.
...
+ 
+Boolean
+ - Booleans must be converted to "true" or "false"
+ - Representation is case sensitive, e.g. "TRUE" or "FALSE" are not valid payloads.
+ - An empty string ("") is not a valid payload
+ 
+Enum
+ - Enum payloads must be one of the values specified in the format definition of the property
+ - Enum payloads are case sensitive, e.g. "Car" will not match a format definition of "car"
+ - Enum matches are exact, e.g. A payload of "car ", will not match a format definition of "car". " car", "car" and "car " Represent three valid and distinct enum values
```
boc-tothefuture: What is the consensus on this?  This was my understanding of the spec as it is written today.  Do we want to allow spaces in enums? Do we want to support leading and trailing whitespace and have those be different enum values?

boc-tothefuture:
Are devices/controllers prohibited from sending numbers outside of the ranges specified in the format for integer and float?

boc-tothefuture:
@marvinroger - can you take a look at this?

convention.md
davidgraeff: Most of the time people will use operating system or language based string splitters and those tend to ignore whitespace. I would say it is undefined behaviour if whitespace is included but wouldn't make it "distinct enum values"

convention.md
davidgraeff: I will change the openHAB implementation. Less data on the wire without whitespace

davidgraeff (approved):

davidgraeff:
@boc-tothefuture Is this still WIP or can we merge what you have done so far and refine everything in later PRs if required?

Thalhammer (approved):

convention.md
boc-tothefuture: Ok, I will update it to say that they aren't distinct enum value..
What languages with splitters by default strip whitespace?  Java doesn't, kotlin doesn't, ruby doesn't...

boc-tothefuture:
@davidgraeff Are we OK with the definition of a float?  I know by default it will break java and kotlin, no idea about python, ruby, etc.

```
public static void main(String[] args)
  {
     System.out.println(Double.MAX_VALUE);
 
  } 
```
` 1.7976931348623157E308 `

convention.md
boc-tothefuture: In my opinion we shouldn't allow trailing or leading whitespace for enums.  I believe both the device and controller should trim all enum values.  It seems like supporting it would lead to errors with confusing troubleshooting with little to no gain.

davidgraeff: C and C++ do :)

davidgraeff: The embedded device decides how it publishes its enums. if without whitespace is the better option all right. But the controller should support both. A `.split(",")` and then access by get(n).trimmed() should work in all languages.

davidgraeff:
Allow floating point numbers up to +1.79×10^308. I'd say we use the most conservative limit here so that it works with most languages

convention.md
Thalhammer: > C and C++ do :)

As far as I know, neither C nor C++ trims whitespaces if you split a string. But maybe I'm missing some API that does.
That aside, I think that whitespaces should not make a difference for matching. I.e. "car" = " car" = "car ".

boc-tothefuture: I will update the enum section to capture the feedback in this discussion.

boc-tothefuture:
What is the conservative limit for floating point numbers? 
In Java:
  1,000,000 is output as 1000000.0
  10,000,000 is output as 1.0E7

10 million doesn't seem unreasonable for a floating point number.   Will other common languages - C, C++, C#, Python, Ruby parse the exponent notation without undue burden on the developer?

Looking at other serialization systems  that are supposed to work cross-language (Avro for example), convert floating point numbers to bits. Not suggesting we do that here, just a data point.

Thalhammer:
Isn't the exponent notation disallowed by the spec?
In the last version that I read it was explicitly noted as an invalid example.
But yes, most languages will be able to parse an E notation, even though some really small micros might need an implementation by the developer.

davidgraeff:
Maybe just put a note in the text saying that the exponential notation could lead to incompatibilities and should be avoided.

boc-tothefuture:
@Thalhammer I disallowed it from the spec based on my interpretation of the intent of the protocol.  I am not sure my interpretation is valid.  
I am OK with @davidgraeff proposal to allow it and put a note in place about avoid exponent notation where possible if others agree.

ThomDietrich:
A few general Markdown comments:

* The single words in front of the itemizations might look ugly in the resulting document (depends on the renderer).
* Please check your indention. Itemization `-` should not have a preceding space
* Please escape the multiplication sign under floats as `&ast;`

Also: Now that the payload chapter became quite long and more specific I wonder if this information would be better suited at another location of the document (e.g. further down).

convention.md
```diff
@@ -18,11 +18,49 @@ The underscore (`_`) is used and reserved for Homie *node arrays*.
 
 ### Payload
 
-Every MQTT message payload MUST be sent as UTF-8 encoded string.
-A numeric data type MUST be converted to a literal string with the dot character as the decimal separator for the `float` data type.
-An empty message is not valid as payload for numeric data types.
-The value published as payload MUST be valid for the respective property/attribute type as per the convention definition below.
-Booleans MUST be converted to "true" or "false".
+ - Every MQTT message payload MUST be sent as a UTF-8 encoded string
+-  The value published as payload MUST be valid for the respective property/attribute type as per the defined convention
```
ThomDietrich: > as per the list below

?

convention.md
```diff
@@ -18,11 +18,49 @@ The underscore (`_`) is used and reserved for Homie *node arrays*.
...
+ 
+String
+ - String types are limited to 268,435,456 characters 
+ - An empty string ("") is a valid payload
+ 
+Integer
+ - Integer types are UTF-8 encoded representations of 64-bit signed whole numbers
```
ThomDietrich: The previous state was:

> A numeric data type MUST be converted to a literal string with the dot character as the decimal separator for the `float` data type.

The first part of this is not equally good represented here. "UTF-8 encoded representations" does not imply "converted to a literal string"

convention.md
```diff
@@ -18,11 +18,49 @@ The underscore (`_`) is used and reserved for Homie *node arrays*.
...
+ - Integers range from -9,223,372,036,854,775,808 (-2<sup>63</sup>) to 9,223,372,036,854,775,807 (2<sup>63</sup>-1)
+ - The payload may only contain whole numbers and the negation character "-". No other characters including spaces (" ") are permitted 
+ - A string with just a negation sign ("-") is not a valid payload
+ - An empty string ("") is not a valid payload
+ 
+Float
+ - Float types are UTF-8 encoded representations of 64-bit signed floating point numbers
```
ThomDietrich: See above

convention.md
```diff
@@ -18,11 +18,49 @@ The underscore (`_`) is used and reserved for Homie *node arrays*.
...
+ - Floats range from 2<sup>-1074</sup> to (2-2<sup>-52</sup>)*2<sup>1023</sup>
+ - The payload may only contain whole numbers, the negation character "-", and the decimal separator ".", no other characters, including spaces (" ") are permitted 
+ - The dot character (".") is the decimal separator (used if necessary) and may only have a single instance present in the payload
+ - Representations of numeric concepts such as "NaN" (Not a Number) and "Infinity" are not a valid payload
```
ThomDietrich: I agree but I wonder if we should shortly justify this decision

convention.md
```diff
@@ -18,11 +18,49 @@ The underscore (`_`) is used and reserved for Homie *node arrays*.
...
+ - Representations using exponent notation ("e" or "E") are not valid payloads
+ - A string with just a negation sign ("-") is not a valid payload
+ - An empty string ("") is not a valid payload
+ 
+Boolean
+ - Booleans must be converted to "true" or "false"
```
ThomDietrich: > ... to the literal strings...

convention.md
boc-tothefuture: My assumption is that they are not broadly supported, or if they are, they are implemented differently across different platforms.

- Is that assumption valid?
- Is that enough to justify exclusion?

ThomDietrich: I'm not too certain about that. NaN and Inf are supported by many implementations (even though always slightly different).
My reasoning comes ones again from the convention and is basically the same as the one for empty strings: A property represents a real *property* of a device, e.g. a temperature. The device is in charge to publish a valid current value regularly (forming a time series of measurements). Non-real values would not make sense in this context.

The case of a property with non-continuous state (i.e. without continuous update) was discussed in other issues and relies on either the stateless concept, another property (e.g.` TempMeasurementActive`), or a user-defined logic (not preferred but of course an option).

Do others agree?

boc-tothefuture:
> Maybe just put a note in the text saying that the exponential notation could lead to incompatibilities and should be avoided.

I tested exponent notation on python, ruby, go, java and C and they all seem to have no issues parsing it.  I removed the restriction on exponent notation from the spec.

boc-tothefuture:
> * The single words in front of the itemizations might look ugly in the resulting document (depends on the renderer).
@ThomDietrich  What are you suggesting we do here instead?

> Also: Now that the payload chapter became quite long and more specific I wonder if this information would be better suited at another location of the document (e.g. further down).

I agree @davidgraeff should this be moved down to the properties section?

convention.md
boc-tothefuture: My inclination is to not justify the exclusion of NaN or Infinity.  I am not sure how to concisely articulate the reasoning and I think people may use homie for things that don't strictly adhere to the definition above of a "real property".

Do you have a suggestions for a concise justification?

convention.md
boc-tothefuture: Fixed

convention.md
boc-tothefuture: Fixed

convention.md
boc-tothefuture: Fixed

davidgraeff:
It is quite a huge section now, I'd like it to stay as preface section. It will also cause trouble with the website generator, which extracts specific sections to generate a document preface that is used for all versions of the specification.

boc-tothefuture:
BTW @davidgraeff the new website looks awesome.

timpur (approved):

davidgraeff (approved):

PR #138, opened 2018-11-20
Last will limitation

### LINKED ISSUE #136: Last will topic can only be set once per MQTT client/ represent a single Homie device
This may be obvious, but I've only discovered it just now: using a single MQTT client for publishing multiple homie devices (or other non-homie use) is not possible since last will can only be set once and for a single topic as far as I understand. This seems to be an MQTT limitation.

Is it expected to use one MQTT client per published device? It seems like a waste of resources?

davidgraeff:
Only one last will possible per MQTT connection. Publishing multiple Homie devices requires multiple connections therefore.
Fixes #136


davidgraeff:
@Thalhammer @timpur

Thalhammer (approved):

PR #140, opened 2018-11-23
Remove requirement for the homie/ base topic.

davidgraeff:
The requirement of having a "homie/" base topic is actually not necessary
and prevents adoption in commercial products. Enforce how controllers have
to perform auto discovery instead.

### DIFF
--- a/convention.md
+++ b/convention.md
@@ -74,10 +74,12 @@ As a consequence a new MQTT connection to the brocker is required per published 
 
 ## Base Topic
 
-The MQTT base topic you will see in the following convention will be `homie/`.
-If this base topic does not suit your needs (in case of, e.g., a public broker), you can choose another.
+The root topic in this document is `homie/`.
+If this root topic does not suit your needs (in case of, e.g., a public broker or because of branding),
+you can choose another.
 
-Be aware, that only the default base topic `homie/` is eligible for automatic discovery by third party controllers.
+Homie controllers must by default perform auto-discovery on the wildcard topic "+/+/$homie".
+Controllers are free to restrict discovery to a specific root topic, configurable by the user.
 
 ## Timings
 


davidgraeff:
@Thalhammer

convention.md
```diff
@@ -74,10 +74,12 @@ As a consequence a new MQTT connection to the brocker is required per published
...
 
-The MQTT base topic you will see in the following convention will be `homie/`.
-If this base topic does not suit your needs (in case of, e.g., a public broker), you can choose another.
+The root topic in this document is `homie/`.
+If this root topic does not suit your needs (in case of, e.g., a public broker or because of branding),
+you can choose another.
 
-Be aware, that only the default base topic `homie/` is eligible for automatic discovery by third party controllers.
+Homie controllers are required to perform auto-discovery on the wildcard topic "+/+/$homie"
+to stay independant of the choosen root topic.
```
timpur: Why not make it even more open, `{base_topic}/+/$homie`

Cant `homie/location1/` also be a base topic ? The controller would only need to know a list of base topics to search ?

timpur: default would be `homie/`

Thalhammer: The problem is that you can't subscribe to `#/$homie`, because # is only supported as the last character.

Thalhammer (approved):
Seems right to me

convention.md
timpur: so the default would be `+/+/$homie` but i still think controller should allow you to add your own base paths to search on ?

Id recommend what @davidgraeff has said but not limit to only this ? That's all im trying to say??

davidgraeff: Controllers are free to show all or just a selection of discovered homie devices. That implies that they can limit the subscription to a specific root topic. Why should this be extra mentioned? The document should be as concise as possible, I think.

fermuch:
@davidgraeff what is the reasoning behind this change?

As far as I understand, the base topic is used for separating concerns and not (only) for branding. For example: I don't want my private sensors to show under the public (`homie/`) space.

One trivial example:
* Let's assume we have only one LAN network, with multiple controllers (each doing a different task).
* The general controller subscribes to `homie/+/$homie`, and devices under this topic are "public".
* Now, a more specific controller only subscribes to `heater/+/$homie`

The general controller shows an interface to update values on all settable properties, and this is good.
The heater controller is scoped for only certain kind of devices, which should not be updated from outside of the controller (think of it like a VLAN)


By using a global handler in the public controller (`+/+/$homie`) I can't separate my devices under the same network anymore, because there's no way in homie to identify who sends a message (is it coming from the public controller or my controller?) or to specify if a device should be registered in the controller.

davidgraeff:
@fermuch If you read this sentence: "Be aware, that only the default base topic `homie/` is eligible for automatic discovery by third party controllers."

Don't you feel the need to use homie/ as root topic to be discoverable by 3rd party homie controllers?
This is bad, if we want to convince manufacturers like Shelly Inc to use Homie. Basically all manufacturers have their name or product name as root topic (e.g. shelly/ in the aforementioned case).

> I don't want my private sensors to show under the public (homie/) space.

Homie is made for auto discovery. There is no such thing as "private" sensors. A controller should in its default settings, discover every available homie device.

Using the base topic for scoping wasn't the intention, You can use a `mybasetopic/device123/scope/heating`, `mybasetopic/device123/scope/sensors` topic and have your controller subscribe to `+/+/scope/heating` etc.

Thalhammer:
Maybe it would be best to note that controllers are required to do discovery at least on `homie/#` but are free to include additional topic in auto discovery ?

davidgraeff:
@Thalhammer  But additional root topics are not known beforehand.

davidgraeff:
I have removed the word "required" which was the main criticism I think

fermuch:
> I have removed the word "required" which was the main criticism I think

I have no criticism now. The only part bothering me was requiring it.

> Don't you feel the need to use homie/ as root topic to be discoverable by 3rd party homie controllers?
This is bad, if we want to convince manufacturers like Shelly Inc to use Homie. Basically all manufacturers have their name or product name as root topic (e.g. shelly/ in the aforementioned case).

By using `homie/`, I've always thought of it as "I want this device to be part of a global network of homie devices", and by using anything else, I've assumed "I want to handle things on my own".

All of the controllers I've seen on github have `homie/` as a default, but can be configured to use any other id. This is an explicit change you need to do, and you're manually changing scopes.

I've never seen a controller using `+/+/$homie` for discovery (but I might be wrong!) and I fail to see how it is useful. I understand how it might be required for branding (as in your example), but then, a controller should be explicitly told to adopt that base topic instead of just grabbing every topic it can find, or at least that's how I see it.

Thalhammer:
> @Thalhammer But additional root topics are not known beforehand.

No they are not but this is not required at all. Until now a controller was not allowed (by the convention) to do discovery on any topic except `homie/`. But as a vendor I'd like to do discovery and still have my devices in a different root. Until now this would have violated the convention.

I'm fine with the way it's worded now.

davidgraeff:
> All of the controllers I've seen on github have homie/ as a default, but can be configured to use any other id. 

> I've never seen a controller using +/+/$homie for discovery (but I might be wrong!) 

This is because 3.x of the homie conventions requires this to be useful aka discovery works out of the box. When implementations get updated to 4.0 you will hopefully see that controllers find as many homie devices on the broker as possible.

> and I fail to see how it is useful. I understand how it might be required for branding (as in your example), but then, a controller should be explicitly told to adopt that base topic instead of just grabbing every topic it can find, or at least that's how I see it.

That is the point. In the default configuration of a controller ALL homie devices should be found. That is the entire purpose of this convention to allow discoverability. The user can intervene to restrict the discover topic if he wants to.

ThomDietrich:
Only seeing this change now. Excellent improvement for broad adoption

PR #143, opened 2018-11-26
Remove $stats

davidgraeff:
Should be modeled as an extension.
Adding some but not all and every $stat (who decides what to include?) is half baked.
And $stats violates the idea of using the dollar sign for attributes only.


timpur (changes requested):
@davidgraeff, you forgot to remove from device ? `$stats`

timpur:
Should this just be a node ?

timpur (approved):

Thalhammer (approved):

PR #147, opened 2018-11-30
Fix extension suffix typo

ThomDietrich:
Do not merge jet. What's the message here? Are we disallowing the project name: https://the-most-amazing-homie-extension.com?

### DIFF
--- a/convention.md
+++ b/convention.md
@@ -328,6 +328,6 @@ A Homie device may therefore support extensions, defined in separate documents.
 Every extension is identified by a unique ID and will be linked from this section.
 
 The ID consists of the reverse domain name and a freely chosen suffix.
-The domain and suffix `homie` is reserved.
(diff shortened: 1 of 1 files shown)


davidgraeff:
"example.org.homie" is disallowed (suffix may not be "homie" on its own)
"homie.org.my-extension" is disallowed (domain may not me homie)

davidgraeff:
Any progress on this front? :)

ThomDietrich:
I am unsure if we should and can forbid "homie" in the domain name...

davidgraeff:
At some point a homie domain might be registered for this convention. I just want to avoid the situation that someone has extensions with ids like "org.homie.extname" or whatever the official domain would be.

ThomDietrich:
Good idea I guess. Okay let's have it in the convention.

PR #155, opened 2019-01-20
Remove timings paragraph. Rename behaviour to lifecycle.

davidgraeff:
A controller need a way to figure out when to assume default values for optional topics.

Instead of relying on a timing, the $state attribute should be used.
A next question would be if we allow topic changes ($nodes, $properties) while in $state=ready mode.

### DIFF
--- a/convention.md
+++ b/convention.md
@@ -87,12 +87,6 @@ you can choose another.
 Homie controllers must by default perform auto-discovery on the wildcard topic "+/+/$homie".
 Controllers are free to restrict discovery to a specific root topic, configurable by the user.
 
-## Timings
-
-As soon as a device starts to publish any Homie related topic,
-it MUST finish with all topics within a timeframe of 500ms.
-Controllers should assume the default for topic values not received within this timeframe.
-
 ## Topology
 
 **Devices:**
@@ -136,7 +130,7 @@ The following device attributes are mandatory and MUST be send, even if it is ju
 |-------------|--------------------------------------------------------------------------:|
 | $homie      | The implemented Homie convention version                                  |
 | $name       | Friendly name of the device                                               |
-| $state      | See [Device behavior](#device-behavior)                                   |
+| $state      | See [Device Lifecycle](#device-lifecycle)                                   |
 | $nodes      | [Nodes](#nodes) the device exposes, separated by `,` for multiple ones.   |
 | $extensions | Supported extensions, separated by `,` for multiple ones.                 |
 
@@ -156,15 +150,14 @@ homie/super-car/$implementation → "esp8266"
 homie/super-car/$state → "ready"
 ```
 
-#### Device Behavior
+#### Device Lifecycle
 
-The `$state` device attribute represents, as the name suggests, the current state of the device.
+The `$state` device attribute represents the current state of the device.
 There are 6 different states:
 
 * **`init`**: this is the state the device is in when it is connected to the MQTT broker, but has not yet sent all Homie messages and is not yet ready to operate.
 This state is optional, and may be sent if the device takes a long time to initialize, but wishes to announce to consumers that it is coming online. 
-* **`ready`**: this is the state the device is in when it is connected to the MQTT broker, has sent all Homie messages and is ready to operate.
-You have to send this message after all other announcements message have been sent.
+* **`ready`**: this is the state the device is in when it is connected to the MQTT broker, has sent all Homie messages and is ready to operate. A Homie Controller can assume default values for all optional topics.
 * **`disconnected`**: this is the state the device is in when it is cleanly disconnected from the MQTT broker.
 You must send this message before cleanly disconnecting.
 * **`sleeping`**: this is the state the device is in when the device is sleeping.


boc-tothefuture:
How will this work when the controller reconnects?  It may get the topics in any order and may take an arbitrary long duration to process them (depending on system load, number of topics, load on the broker, etc).

euphi (approved):
Yes, timing is unnecessary, due to the $state topic.

For retained messages sent from the broker to a newly subscribed client, timing may matter (because the order of MQTT messages is lost), but this is beyond the influence of the Homie convention.

davidgraeff:
Yeah I know. This only solves the situation for slow Homie peer devices.
A second lifecycle / state-machine for the communication with the MQTT broker is required for a Homie controller. As long as the broker is keeping sending stuff, a controller can assume a working connection and otherwise a timeout for that connection is somehow required or implementation specific.

euphi:
@boc-tothefuture The homie convention can't handle this case.

As a workaround, a newly subscribed client can send a not-retained message to the broker to a MQTT topic where it is subscribed to.
*Usually* the broker will queue this message and send it after the old stored (retained) messages.

However, maybe the problem is not that relevant in practice:
A controller that wants to learn the advertised devices / nodes and their properties, just can update its list "in realtime".

Thalhammer:
How about building some kind of list of published topics the way we list nodes and properties?
This way a controller would exactly know what to expect and when everything arrived.
It also should be easy to implement on devices and controllers.

convention.md
```diff
@@ -156,15 +150,14 @@ homie/super-car/$implementation → "esp8266"
...
 
-The `$state` device attribute represents, as the name suggests, the current state of the device.
+The `$state` device attribute represents the current state of the device.
 There are 6 different states:
 
 * **`init`**: this is the state the device is in when it is connected to the MQTT broker, but has not yet sent all Homie messages and is not yet ready to operate.
 This state is optional, and may be sent if the device takes a long time to initialize, but wishes to announce to consumers that it is coming online. 
-* **`ready`**: this is the state the device is in when it is connected to the MQTT broker, has sent all Homie messages and is ready to operate.
-You have to send this message after all other announcements message have been sent.
+* **`ready`**: this is the state the device is in when it is connected to the MQTT broker, has sent all Homie messages and is ready to operate. A Homie Controller can assume default values for all optional topics.
```
ThomDietrich, line 160: Please break one sentence per line

ThomDietrich (approved):
Looks good to me.

We have to differentiate between a connecting client and the discovery by a controller at any later point.

The first aspect is covered by `$state=ready` and the second needs to be handled correctly by the controller and relies on the responsiveness of the broker. The controller might need to wait for all retained topics. That's probably easy to implement and I guess extensively tested by @davidgraeff.

@davidgraeff wdyt, should this be discussed in the controller developer guidelines (#51)?

timpur (approved):

ThomDietrich:
> @davidgraeff wdyt, should this be discussed in the controller developer guidelines?

Also, did you miss my comment?

davidgraeff:
> Also, did you miss my comment?

No I didn't. We don't have such a section yet though. Do you aim for the website or the convention document itself?

ThomDietrich:
I was refering to this comment https://github.com/homieiot/convention/pull/155#pullrequestreview-194413594

> Please break one sentence per line

----

Regarding the guideline: All the more important that we capture the ideas for content of this section. This is where I am confused because you've opened and closed a few issues and now there is non open any more 😕 
I also do not understand your question. Shouldn't the website repository be a place of presentation and style but all convention related information and knowledge belongs in the convention repository!? Come to think of it: Why are there two repositories to begin with?

davidgraeff:
> Why are there two repositories to begin with?

Because martin roger doesn't want to give me full access to this repo, but for a website deployment environment I required this. So we have two repos now ^^.

> Shouldn't the website repository be a place of presentation

But also about implementation details, I think. The specification should be a clean specification only.

ThomDietrich:
Things like this shouldn't hold us back. @marvinroger would you agree that active contributors should have the correct permissions to drive this project forward. Please consider entrusting one or two of the currently active developers. Jm2c 😉 

Two repos problem aside: Everything related to the convention should be in the repository. Be it the main document, example implementations, FAQs or guidelines. All of those belong together and form the complete picture of the convention. If at all possible I'd strive to merge the two repositories.

euphi:
> Things like this shouldn't hold us back. @marvinroger would you agree that active contributors should have the correct permissions to drive this project forward. Please consider entrusting one or two of the currently active developers. Jm2c 

@marvinroger gave me access to homie-esp8266 today, so I can help in progress V3 implementation for ESP8266.

@marvinroger: For ensuring project progress, I think it would be best to give some few people access to the homieiot organization (make them member), so they can control rights for the projects repos.

marvinroger:
@davidgraeff @ThomDietrich @euphi @timpur alright, you now have full access to the homieiot organization.

@davidgraeff just to be clear, I never intended "not to give you full access to this repo". I am just overloaded with a lot of personal matters, and I don't have time to read GitHub, so I might have misunderstood a few things.

Finally, I apologize if I slowed you, and the different Homie projects, down. I am infinitely grateful to all of you who keep this project alive.

OSS love, if that's a thing 😄❤️

davidgraeff:
@ThomDietrich I'd be fine with a merge of the homepage and convention repos. We can even build the website on on separate branch of this repo and remove the third homepage-build-result repo as well, if you like.

ThomDietrich:
@marvinroger thank you! With this step we can now continue to push the success of the homie convention. Hope to see you again soon!

@davidgraeff my opinion is that this would be the best way forward. @euphi wdyt?

euphi:
I didn't had a look into doc generation yet, so I don't have a well-founded opinion about this.

Regarding the homepage, I think it should also give some information about implementations (like homie-esp8266) , integrations (like the openhab MQTT binding) and Tools (like the OTA scripts).

If you think, this can be handled together with the Convention Repo, just merge them :blush:

PR #157, opened 2019-02-28
Point Last Will section to `$state`

ibloat:
`$ready` does no longer seem to be a thing,
Lifecycle is tracked in the `$state` attribute so the last will should update that instead

### DIFF
--- a/convention.md
+++ b/convention.md
@@ -75,7 +75,7 @@ All messages MUST be sent as **retained**, UNLESS stated otherwise.
 ### Last will
 
 MQTT only allows one last will message per connection.
-Homie requires a last will for the `homie` / `device ID` / `$ready` attribute, see [Device Behavior](#device-behavior).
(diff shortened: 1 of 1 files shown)


davidgraeff (approved):
True. Thanks

euphi (approved):

euphi:
@davidgraeff I think this should be tagged as 3.0.2?

davidgraeff:
We would need to backport that for a revision update. The current master contains breaking changes and will be 4.0

euphi:
Ah yes, I forgot. Luckily, the v3.0.1 tag does not contain this error, because it only defines the possible state for $state:

> 
> The $state device attribute represents, as the name suggests, the current state of the device. There are 6 different states:
> 
> [....]
> - lost: this is the state the device is in when the device has been "badly" disconnected. You must define this message as LWT.
>

Issue #170, opened 2019-07-06
How to define/propose an extension? Concret: openHab label extension


zocker1996:
Is their a guide how to define an extension?
What is an extension allowed to do?

Let me give you a concete example:
I want to propose an extension, which allowes you to add openHab labels to nodes an properties. These labels can then automatically be processed by openHab and linked to the appropriate openHab items (for properties) / channel groups (for nodes). A simple way to add such labels would be to just create a new attribute "$openhab-label" for properties and nodes. For example:
homie/super-car/engine/temperature/$openhab-label

What steps should I take to make an extension convention, what kind of information do I need to write down, is there a template?

davidgraeff:
This is a primer, so there is no template yet. I would expect $type definitions and predefined properties and attribute values in such an extension. Maybe you give it a go as a PR to this repo? (Separate document)

If possible, make it independent of openHAB and think about the problem space and what you actually want to archive.

zocker1996:
I'll try to come up with an idea for a template and a general structure based on the original convention document.

Issue #102, opened 2018-05-06
Homie-Statistical-Values


timpur:
There has been talk if we should implement device stats as a node. The Mian reason (IMHO) is so that the user can add there own device sensor out side of the spec, and these stats would automatically be discoverable. 

We could leave the spec as is, recommending the most common stats, but also recommend implentations implent stats as a node if they like. 

I'm not sure of the solution, but I like the idea, as I don't want to be adding every possible stat to the convention.... (Over time)

ThomDietrich:
Hey,
could you please summarize what counts as "device stats" and why we need it? Why is an extension needed?

davidgraeff:
Any user defined (probably read-only) property serves as a "statistic value". It probably makes sense to define some "$type"s instead like "battery" etc.

timpur:
So currently in 3.0 we have a section defined as `Device Statistics`, this included stats like:
* $stats/signal
* $stats/cputemp
* $stats/freeheap
* ...


The spec, doesn't specify if this is or isnt open to users or implementation to add there own device stat or not. The issue is, if it is locked down then over time will have request to add more and more stats... and if it is open for user or implementation to add there own, then how do we auto discover these dynamic stats? My proposal is to just have a these implemented as a node, which comes with property discoverabilty, with unit and format ect....

davidgraeff:
To resurrect this old topic: Is everybody d'accord to remove $stats for 4.0 and reintroduce it in a different form (probably via extension?)?

timpur:
@davidgraeff sounds good, lets come up with an extension

arjanmels:
I think the arguments to limit the $stats section are sound.
One feature that will also be lost however, is the heartbeat (removal of $stats/interval $stats/uptime).

How will online/offline detection be handled going forward?

davidgraeff:
> How will online/offline detection be handled going forward?

MQTT last will.

arjanmels:
@davidgraeff Thanks, that is indeed a clean solution.

EP-u-NW:
With the specification of the stats extension in #171, what further steps need to be taken to close this issue?

davidgraeff:
@timpur Please reopen if necessary

PR #171, opened 2019-07-10
Extension Convention

ghost:
As described in [issue 170](https://github.com/homieiot/convention/issues/170) I tried to come up with an idea, what an extension is and what it is allowed to do. This ideas are bundled in the *homie extension convention.md* document and need to be discussed.

As a second step I wrote a template document for extensions based on the extension principles (see file *extension template.md*). Again, as the extension principles change, this document should be revised.

Subsequently, I wrote the extension document for the concrete extension I want to realize (*homie meta extension.md*). 

After that I though it would be nice to have a table listing all extensions, so I added *known extensions.md*.

Then I remembered [issue 102](https://github.com/homieiot/convention/issues/102) and thought it would be nice to try to realize a solution based on the extension template. As result, this pull request contains *homie legacy stats extension.md* and *homie legacy firmware extension.md*. Note that I imagined that the last two extensions could somehow be *official* extensions, so I chose *The Homie Community* as author and also used the reserved term *homie* in the extension ID.

I suggest to move all extension related things to a new branch, but it seems you can not do that using pull requests, so I decided to create a new folder instead.

Maybe all this is a bit much for a single pull request, so take your time reading it and happy discussion :)


davidgraeff:
Pretty slick. Thanks for the PR.
I have two general comments:
* Please do not use whitespace in filenames. Replace with dashes or underscores.
* "homie extension convention.md" should not be part of this PR IMO. It's more of an instructional text that should go to the website repo or party be integrated into the convention itself.

I'm happy with the template document and vote for a fast merge and revise later if necessary. All 3 extensions look complete.

extensions/documents/homie meta extension.md
```diff
@@ -0,0 +1,121 @@
...
+
+Version: **<!--VERSION-->1.0.0<!--VERSION-->**
+Date: **<!--DATE-->10. Jul 2019<!--DATE-->**
+Authors: **<!--AUTHORS-->[EPNW](https://epnw.eu)<!--AUTHORS-->**
+License: **<!--LICENSE-->[CCA 4.0](https://homieiot.github.io/license)<!--LICENSE-->**
+
+## Abstract
+This extension defines how to add metadata and tags to devices, nodes and properties.
+
+Tags are simple annotations that ever device, node oder property can have.
```
davidgraeff: ever -> every

extensions/documents/homie meta extension.md
```diff
@@ -0,0 +1,121 @@
...
+Each main-key-value-pair may have nested sub-key-value-pairs.
+Having metadata might be useful for Homie controllers.
+A concrete usecase is the openHAB controller, which relies on metadata and tags to make voice assistants like Amazons Alexa or the Google Assistant able to control it.
+With this extension, properties are able to advertise how they want to be treated.
+
+## Extension Identifier
+The ID of this extension is `eu.epnw.meta`.
+
+## Homie Version
+This extension supports Homie `3.0.1` and `4.0`.
```
davidgraeff: Homie versions are semantic. Better use "4.x" instead of "4.0" as this extension is probably valid for the entire lifetime of Homie 4.

extensions/documents/homie meta extension.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Extension Datatypes
+This extension defines no new datatypes.
+
+## Extension Attributes
+
+### Extension Attributes for any kind of Items
+
+The following attributes are all **optional** and may be added to any kind of item (devices, nodes or properties).
```
davidgraeff: It is not defined if those attributes need to be "retained"

extensions/documents/homie meta extension.md
```diff
@@ -0,0 +1,121 @@
...
+| $tags                | List of *tags* for the item            | Comma (`,`) separated list of *tags* |              |
+| $meta/$mainkey-count | The number of *mainkeys* this item has | Positive Integer                     | 0            |
+
+
+Since *tags* are represented by a comma separated list, a limitation is that a tag itselfe MUST NOT contain a comma.
+
+**Examples**
+Assuming the base topic is *homie*, device ID is *super-car*, the node is *engine* and the property is *temperature* then:
+```java
+homie/super-car/engine/temperature/$meta/$key-count → "2"
```
davidgraeff: There is a reason why Homie has $properties and $nodes. How do you know if a $meta sub topic is still valid or a left over. Also: A controller never knows how many meta subtopics it will receive without a "list" of them.

extensions/homie extension convention.md
```diff
@@ -0,0 +1,61 @@
...
+Version: **<!--VERSION-->0.1.0<!--VERSION-->**
+Date: **<!--DATE-->10. Jul 2019<!--DATE-->**
+License: **<!--LICENSE-->[CCA 4.0](https://homieiot.github.io/license)<!--LICENSE-->**
+
+## Abstract
+This document specifies how to create an extension documents and defines the responsibilities of extensions.
+For a template to define extensions take a look at the [extension template]().
+
+## Licensing
+Every extension must be published using a license.
```
davidgraeff: This license paragraph should probably go into the main convention (and the website)

extensions/homie extension convention.md
```diff
@@ -0,0 +1,61 @@
+# The Homie Extension Convention
```
davidgraeff: This document should be part of the website and parts of it integrated into the main spec

extensions/homie extension convention.md
```diff
@@ -0,0 +1,61 @@
...
+## Mandatory Version Attributes
+Each extension must have an $version attribute and and $homie-versions attribute.
+An extension is added to a device by adding the extension ID to the [$extension](https://homieiot.github.io/specification/spec-core-develop/#device-attributes) attribute of the device.
+Then the device needs to publish the version of the extension to
+```java
+base topic/device ID/$extension ID/$version
+```
+and a comma (`,`) separated list of the Homie versions the extension supports to
+```java
+base topic/device ID/$extension ID/$homie-versions
```
davidgraeff: I think we should reduce the amount of topics required for an extension. For example by collapsing extension id, version and homie version range into one topic with a natural looking separator like a ";" eg: "my-cool-extension; 1.0; [3.0,5)"

For a controller that doesn't care this is less stressful and an mqtt broker has to retain less topics (faster startup)

extensions/known extensions.md
```diff
@@ -0,0 +1,10 @@
+# Known Extension Documents for Homie
```
davidgraeff: Originally I thought that we include such a list into the main spec. But might indeed be a good idea to just link to this document instead. And the website can fetch this markdown file and render it somewhere additionally.

EP-u-NW:
Thanks for your answer!
I'll comment your reviews.
How would you suggest we should procced after that?
Should I create a new pull request, which only contains the reviewed extensions and the template,
and you will add the parts of extensions convention to the convention (as an ohter pull request)?

extensions/homie extension convention.md
EP-u-NW: That sounds like a good idea. This way, there is no need for a topic ID that violates the convention (by containing dots).

extensions/homie extension convention.md
EP-u-NW: Yes, it should somehow be part of the main spec. But I think we should try avoid getting the main spec as seen on the website to long. As form my personal experience, I decided to take a closer look at Homie because the convention page on the website is relativly short compared to other specifications. What exactly an extension is might not be that interesting for new users who just want to get a quick idea of the convention. Althrough knowing that there are extensions is important ofcourse. So I suggest we should try to make a separate tab on the website.

extensions/documents/homie meta extension.md
EP-u-NW: I stated in the extension convention that every new attribute has to be retained, so per definition, this attributes are retained, too. But since you mention this here, I feel like we should discuss if requiring every extension attribute to be retained is a good idea.

extensions/documents/homie meta extension.md
EP-u-NW: You are right, I haven't thought about that! My first approach was to use a list, and then have the subtopics accessed by their keynames instead of indexes. E.g:
```java
homie/super-car/engine/temperature/$meta/$keys → "alexa,homekit" 
homie/super-car/engine/temperature/$meta/alexa/$value → "Fan" 
```
But then I noticed that this is a very strong limitation, since the keys are then used as topic IDs. And topic IDs "MAY contain lowercase letters from a to z, numbers from 0 to 9 as well as the hyphen character (-)" (quote from the [covention](https://homieiot.github.io/specification/)).
I guess a solution would be to make the indexing still array like, but have a $keys attribute like in the example above. Then
```java
homie/super-car/engine/temperature/$meta/1/$key → "alexa" 
homie/super-car/engine/temperature/$meta/1/$value → "Fan" 
```
is only valid, if the keys list contains `alexa`.

extensions/homie extension convention.md
EP-u-NW: Just to clarify: This paragraph should make sure every extension that is published (and probably added to the known extensions list) clearly states which license applies to it.

I think this notice is part of the instructional text. I agree that if we move all the instructional text to the main spec, we should move this, too. See my other comment to your review of *extensions/homie extension convention.md* about moving the whole instructional text.

extensions/documents/homie meta extension.md
davidgraeff: > keys are then used as topic IDs

That's a reasonable limitation. Remember that you can always base48 or base64 encode a key

extensions/homie extension convention.md
davidgraeff: I'm with you. I do not suggest to put the entire text into the main spec. Only the license part, so one sentence. And the rest goes to the webpage. But nobody will ever go this repository into the documents subdirectory to find "homie extension convention.md" I guess.

davidgraeff:
I suggest you also modify the main convention in this PR, adding the license part to the extensions section and remove the sentence that says that extensions are listed in that section. I will add a link to the website later on, as soon as the website has been updated.

Edit: I have updated the website with an extensions page containing your "extensions convention" text and a short intro. I need to update the build script at some point to generate a list of all extensions in the documents directory and inject that into the extensions page. For now it is a static list.

extensions/documents/homie meta extension.md
EP-u-NW: I redesigned the key addressing. I choose a list as you suggested instead of a count.
But to avoid the limitation regarding the topic IDs, the lists elements are not key names, but just key IDs. The key name and value are then nested topics under the keys topic ID.
I did not like the idea of encoding the key name base64 and use it as topic ID, since then the key name itself is not human readable (and base64 contains upper case letters and `=`,  which are not valid topic characters).

davidgraeff (approved):

davidgraeff:
@Thalhammer @timpur

timpur (approved):
Looks Good To Me
I think this open up some doors in the future
Thanks :)

Thalhammer (approved):

extensions/documents/homie_legacy _stats_extension.md
```diff
@@ -0,0 +1,67 @@
+# Legacy Stats
```
davidgraeff, line 1: I think there is still a whitespace in the name, apart from that all maintainers have approved and we are ready to merge

EP-u-NW:
I would suggest we merge then. Then I can fix the hyperlinks (since I know the path then), and the remaining typos :)

PR #173, opened 2019-07-17
Corections to meta extension

arjanmels:
Correct oder to or and numerals in example to keys .

### DIFF
--- a/extensions/documents/homie_meta_extension.md
+++ b/extensions/documents/homie_meta_extension.md
@@ -58,10 +58,10 @@ For each element in the `$mainkey-ids` list, two nested attributes are **require
 **Examples**
 With respect to the previous example:
 ```java
-homie/super-car/engine/temperature/$meta/alexa/$key → "HomeKit"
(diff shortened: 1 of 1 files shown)


ghost (approved):

ghost (changes requested):
You are right, there should be no numbers.
But in your correction, you assigned the topic id "alexa" to the "HomeKit" key. While this is technically allowed according to the extension specification, I actually intended to assign the "Alexa" key value to the "alexa" topic, and the "HomeKit" key value to the "homekit" topic.

Edit: So what i mean is: you changed the numeral "0" to "alexa", while it would me more appropriate to change "0" to "homekit" and "1" to "alexa".

arjanmels:
Sory, my mistake

EP-u-NW:
Can you make an other commit to fix this :) ?

arjanmels:
Corrected.

ghost (approved):

davidgraeff (approved):

PR #177, opened 2019-08-27
Remove array brackets in super-car example

stefan-muc:
Homie 4.0 dropped array support, so super-car example shouldn't include array brackets


davidgraeff (approved):
True. Thanks

hachi:
@davidgraeff why is this change not reflected on the main site?

davidgraeff:
Is it not? Could you point to the site please?

hachi:
I think it's your canonical site: https://homieiot.github.io/specification/#device-attributes

Issue #191, opened 2020-01-24
Extension proposal: last update time for property


maxberger:
Hi,

I would like to propose an extension which adds a time for a last updated of a property. But there are several items I would like to discuss.

Basic idea: Any property can have an "last updated" value which represents the time at which the backend for this value has actually been updated.

Why is this needed: Some backends, specifically gateways to other systems, may publish "old" values, which may or may not have changed since. This needs to be on a property basis, so adding this to the overall device is not sufficient.

Open Questions:
1 What is the best format for such a timestamp? I see several options:
1.1 seconds since the actual update, integer format
1.2 unix timestamp of the last update, integer format
1.3 time of the last update, some common time format (such as ISO time)

2 What name would fit best? $last-update?

3 What is the best level for this extension? I would need it for property, but is there a reason to also have this on node?

Some Pro's and Con's I see for 1:
1.1: Pro: Very easy to create, does not depend on time synchronisation
1.1: Con: Clients need to calculate timestamp based on their own clock when they receive the message. If the message is RETAINed, that does not work.
1.2: Pro: This is the datatype i get from the backend and which I want to see in the frontend. So in my case, no calculation / conversion is necessary
1.1, 1.2. Pro: No additional datatype needed
1.2, 1.3 Con: Time may not be synchronous between devices, a device may not even have an absolute clock.
1.3: Pro: Human readable
1.3: Con: ISO time is not an current data format, so that would also needed to be added
1.3 Con: The usual time zone issues

What I plan to do:
* attach an extension pull request, once the discussion is done.
* provide a pull request for https://github.com/mjcumming/homie4 with a sample implementation

[Background: I am currently writing a zway -> Homie gateway, hopefully to be published publically at some point].

jamesmyatt:
In my opinion, if you have information like this, then it needs to be atomic (i.e. you get both value and timestamp or neither), especially since you can't guarantee the order that the messages are received or the time between the messages being received, or even that the requested QoS is granted by the broker and that the message arrives at all. For MQTT, this means publishing a single structured message that includes the time and value(s), e.g. JSON-encoded, and relying on the consumers to unpack it.

UNIX timestamps in UTC would be by far the easiest to implement and process. ISO-8601-format times are good for humans but are an unnecessary overhead on resource-limited devices, i.e. the sort of device that you'd want to use MQTT with.

[...]

PR #192, opened 2020-01-25
Added initial version of last-updated extension

maxberger:
See https://github.com/homieiot/convention/issues/191 for discussion.

### DIFF
--- a/extensions/documents/homie_last_updated_extension.md
+++ b/extensions/documents/homie_last_updated_extension.md
@@ -1,20 +1,32 @@
 # Last-updated timestamp
 
 Version: **<!--VERSION-->1.0.0<!--VERSION-->**
-Date: **<!--DATE-->25. Jan 2020<!--DATE-->**
+Date: **<!--DATE-->17. Feb 2020<!--DATE-->**
 Authors: **<!--AUTHORS-->Max Berger<!--AUTHORS-->**
 License: **<!--LICENSE-->[CCA 4.0](https://homieiot.github.io/license)<!--LICENSE-->**
 
 ## Abstract
 This extension adds best-effort last-updated timestamps to properties.
 
-Homie devices will publish a last-updated timestamp along with any property value changes. The last-update timestamp can be used by devices or gateways that do not constantly pull data from the real back-end. In this case, they can report when they last actually updated the data, so that clients can act accordingly, and maybe ignore old data, or display this information in an UI.
+Homie devices will publish a last-updated timestamp along with any property value changes. The last-update timestamp can be sent by devices or gateways that do not constantly pull data from the real back-end. In this case, they can report when they last actually updated the data, so that controllers can act accordingly, and maybe ignore old data, or display this information in an UI.
+
+A common use case for this is a gateway software which publishes device data in Homie format, but actually uses external sensors to get the actual data. Gateways may re-publish the values the gather from the backend, even if these values are older. Using the last-update timestamp, a controller is then able to distinguish between actual value updates, and value updates which are just a republication of older values. 
+
 ## Homie Version
 This extension supports Homie `3.0.1` and `4.x`.
 
 ## Extension Identifier
-The ID of this extension is `name.berger.max.last-updated`.
-Therefore the **$extensions entry** is `name.berger.max.last-updated:1.0.0:[3.0.1;4.x]`.
+
+There are two extensions in this specification. 
+
+The ID of this first extenstion is `org.homie.ext.last-updated`.
+Therefore the **$extensions entry** is `org.homie.ext.last-updated:1.0.0:[3.0.1;4.x]`.
+
+The ID of this extension is `org.homie.ext.last-value-updated`.
+Therefore the **$extensions entry** is `org.homie.ext.last-value-updated:1.0.0:[3.0.1;4.x]`.
+
+A device may implement zero, one, or both extensions. The extension entries refer to the matching property attributes.
+
 
 ## Extension Datatypes
 This extension defines no new datatypes.
@@ -35,7 +47,7 @@ This extension defines no nested node attributes.
 
 ### Property Attributes
 
-This extension defines one property attribute.
+This extension defines two new property attributes.
 
 #### $last-updated
 
@@ -47,11 +59,11 @@ If its used, it **must** follow the description below:
 |---------------------------------------|---------------------------------------------------|------------------------------------|
 | $last-updated        | Timestamp when the property value was last updated by the Homie device | integer: seconds that have passed since the Unix epoch (00:00:00 UTC on 1 January 1970) OR 0 for unknown |
 
-Device usage:
-* The $last-update **must** always be published immediately before a value update.
+Usage notes:
+* The $last-update **should** always be published immediately before a value update.
 * $last-update **must** be 0 or a positive integer
 * $last-update **must** be in UTC
-* If a device cannot get the actual timestamp for whatever reason (clock is skewed, ntp has not yet finished, etc.) it **may** publish 0
+* If a device cannot get the actual timestamp for whatever reason (clock is skewed, ntp has not yet finished, etc.) it **should** publish 0
 * If a device has published a $last-update in the past and is now unable to determine the last-update timestamp, it **must** publish 0 at least once.
 * the Qos and the Retained flags **must** match the value updates.
 
@@ -62,12 +74,61 @@ homie/super-car/wheels/pressure/$last-updated → 1579965965
 homie/super-car/wheels/pressure/ → 32
 ```
 
-This means that the property pressure was last fecthed from its source at unix timestamp 1579965965 (2020 Jan 25, 16:26:05 UTC, or 2020 Jan 25, 17:26:05 CET)
+This means that the property pressure was last fetched from its source at unix timestamp 1579965965 (2020 Jan 25, 16:26:05 UTC, or 2020 Jan 25, 17:26:05 CET)
 
-**Client considerations**
+```java
+homie/super-car/wheels/pressure/$last-updated → 0
+homie/super-car/wheels/pressure/ → 32
+```
 
-Note that MQTT is not atomic, and does not guarantee any order. This means that even though both the value and the last-update timestamp are published together and in the oder above, the clients may receive these messages in any order, or one of them may be lost. Therefore, **$last-updated** is entirely **best-effort**. It **must** **not** be used for any critical operation.
+This means that the device is usually able to sent timestamps, but is unable to do so at this time for any reason.
 
+
+**Controller considerations**
+
+Note that MQTT is not atomic and does not guarantee any order. This means that even though both the value and the last-update timestamp are published together and in the oder above, the controllers may receive these messages in any order, or one of them may be lost. Therefore, **$last-updated** is entirely **best-effort**. It **must** **not** be used for any critical operation.
+
+The special value of 0 can be used for a controller to distinguish devices that are currently unable to get the time from devices which do not update, or from lost mqtt messages. If a controller stores the last-updated value, the value of 0 should be used to clear the data storage.
+
+The value 0 was chosen so that the last-updated value is always a valid timestamp. Controllers which chose not to treat it specifically will still have a usable value.
+
+#### $last-value-updated
+
+The **$last-value-updated** property attribute is **optional**.
+
+If its used, it **must** follow the description below:
+
+| Topic                                 | Description                                       | Payload type                       |
+|---------------------------------------|---------------------------------------------------|------------------------------------|
+| $last-value-updated        | Value and Timestamp, seperated by semicolon | value-type;interger |
+
+Usage notes:
+* the rules for the timestamp are the same as for $last-updated above
+* a device *should* always publish both the actual value and the $last-update-value quickly one after another.
+* if a device has published $last-value-updated, it *must* keep publishing it, with a timestamp of 0 if needed
+* the special timestamp of 0 follows the same rules as for $last-updated above
+
+**Examples**
+Assuming the base topic is *homie*, device ID is *super-car*, node is *wheels*, and the property is *pressure* then:
+```java
+homie/super-car/wheels/pressure/$last-value-updated → 32;1579965965
+homie/super-car/wheels/pressure/ → 32
+```
+
+This means that the property pressure was last fetched from its source at unix timestamp 1579965965 (2020 Jan 25, 16:26:05 UTC, or 2020 Jan 25, 17:26:05 CET)
+
+```java
+homie/super-car/wheels/pressure/$last-value-updated → 32;0
+homie/super-car/wheels/pressure/ → 32
+```
+
+This means that the device is usually able to sent timestamps, but is unable to do so at this time for any reason.
+
+**Controller considerations**
+
+Controllers supporting the extension should prefer values from $last-value-updated. This means once a property value has been read from $last-value-updated, the controler should use that topic from this point on and ignore the regular value updates.
+
+All considerations for the special value 0 from above still hold.
 
 ## Attribution
 - <sup>\[1\]</sup>: [The Homie Convention](https://homieiot.github.io/specification/#), [CCA 4.0](https://homieiot.github.io/license)
--- a/extensions/documents/homie_last_updated_extension.md
+++ b/extensions/documents/homie_last_updated_extension.md
@@ -13,7 +13,7 @@ Homie devices will publish a last-updated timestamp along with any property valu
 This extension supports Homie `3.0.1` and `4.x`.
 
 ## Extension Identifier
-The ID of this extension is `name.berger.name.last-updated`.
+The ID of this extension is `name.berger.max.last-updated`.
 Therefore the **$extensions entry** is `name.berger.max.last-updated:1.0.0:[3.0.1;4.x]`.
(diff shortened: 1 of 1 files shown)


extensions/documents/homie_last_updated_extension.md
```diff
@@ -0,0 +1,73 @@
...
+License: **<!--LICENSE-->[CCA 4.0](https://homieiot.github.io/license)<!--LICENSE-->**
+
+## Abstract
+This extension adds best-effort last-updated timestamps to properties.
+
+Homie devices will publish a last-updated timestamp along with any property value changes. The last-update timestamp can be used by devices or gateways that do not constantly pull data from the real back-end. In this case, they can report when they last actually updated the data, so that clients can act accordingly, and maybe ignore old data, or display this information in an UI.
+## Homie Version
+This extension supports Homie `3.0.1` and `4.x`.
+
+## Extension Identifier
```
EP-u-NW, line 18: There is something wrong with your extension identifier.  Assuming you intended the ID `name.berger.name.last-updated`, which seems a bit odd to me, the  **$extensions entry** should be `name.berger.name.last-updated:1.0.0:[3.0.1;4.x]`. 

I would suggest using a more convenient ID in domain format, like [your country code].berger.last-updated, but thats up to you :)

ghost (changes requested):

extensions/documents/homie_last_updated_extension.md
maxberger: Apologies, there was indeed a spelling mistake here. It should be name.berger.max in order to follow the convention of matching website-owner to id-owner.  (My personal website is https://max.berger.name/ ).

Of course, we could also use the homie identifier, but then this would seem like an "official" extension. Would that be preferred?

davidgraeff: Very often the domain registrar is actually used to identify a purpose like "org" or "io" for public use / invented by public organisations or individuals vs "com" for commercial invented things or used purposes. That's how I see it used in the Java world at least. The reverse domain doesn't always actually exist.

Because the "name" TLD is not common I wouldn't use it here, as it is hardly identified as being a TLD. I suggest something like "org.berger.last-updated" or similar.

extensions/documents/homie_last_updated_extension.md
```diff
@@ -0,0 +1,73 @@
...
+
+#### $last-updated
+
+The **$last-updated** property attribute is **optional**.
+
+If its used, it **must** follow the description below:
+
+| Topic                                 | Description                                       | Payload type                       |
+|---------------------------------------|---------------------------------------------------|------------------------------------|
+| $last-updated        | Timestamp when the data was last updated from the backend / sensor              | integer: unix timestamp in UTC (seconds since 1/1/70) OR 0 for unknown |
```
ThomDietrich:
```suggestion
| $last-updated        | Timestamp when the property value was last updated by the Homie device | integer: seconds that have passed since the Unix epoch (00:00:00 UTC on 1 January 1970) OR 0 for unknown |
```

to be more precise and use the official Homie terminology

ThomDietrich: Additionally: Why is 0 allowed? The use case is unclear and the meaning ambiguous (it's literally just 00:00:00 UTC on 1 January 1970)

Edit: See my longer comment below.

Also: I'm reluctant whether or not I care about the slight ambiguity. Probably totally fine

extensions/documents/homie_last_updated_extension.md
```diff
@@ -0,0 +1,73 @@
...
+
+Device usage:
```
ThomDietrich: "Device"?

extensions/documents/homie_last_updated_extension.md
```diff
@@ -0,0 +1,73 @@
...
+**Examples**
+Assuming the base topic is *homie*, device ID is *super-car*, node is *wheels*, and the property is *pressure* then:
+```java
+homie/super-car/wheels/pressure/$last-updated → 1579965965
+homie/super-car/wheels/pressure/ → 32
+```
+
+This means that the property pressure was last fecthed from its source at unix timestamp 1579965965 (2020 Jan 25, 16:26:05 UTC, or 2020 Jan 25, 17:26:05 CET)
+
+**Client considerations**
```
ThomDietrich: Who is this directed at? Please align "Client" to the Homie lingo

extensions/documents/homie_last_updated_extension.md
```diff
@@ -0,0 +1,73 @@
...
+If its used, it **must** follow the description below:
+
+| Topic                                 | Description                                       | Payload type                       |
+|---------------------------------------|---------------------------------------------------|------------------------------------|
+| $last-updated        | Timestamp when the data was last updated from the backend / sensor              | integer: unix timestamp in UTC (seconds since 1/1/70) OR 0 for unknown |
+
+Device usage:
+* The $last-update **must** always be published immediately before a value update.
+* $last-update **must** be 0 or a positive integer
+* $last-update **must** be in UTC
```
ThomDietrich, line 65: Are these redundant?

extensions/documents/homie_last_updated_extension.md
ThomDietrich: Also add how to react to value 0 here. That would probably answer my question from above.

extensions/documents/homie_last_updated_extension.md
```diff
@@ -0,0 +1,73 @@
...
+* If a device cannot get the actual timestamp for whatever reason (clock is skewed, ntp has not yet finished, etc.) it **may** publish 0
+* If a device has published a $last-update in the past and is now unable to determine the last-update timestamp, it **must** publish 0 at least once.
```
ThomDietrich, line 67: Is this different to unsetting the topic?

extensions/documents/homie_last_updated_extension.md
```diff
@@ -0,0 +1,73 @@
...
+* the Qos and the Retained flags **must** match the value updates.
+
+**Examples**
+Assuming the base topic is *homie*, device ID is *super-car*, node is *wheels*, and the property is *pressure* then:
+```java
+homie/super-car/wheels/pressure/$last-updated → 1579965965
+homie/super-car/wheels/pressure/ → 32
+```
+
+This means that the property pressure was last fecthed from its source at unix timestamp 1579965965 (2020 Jan 25, 16:26:05 UTC, or 2020 Jan 25, 17:26:05 CET)
```
ThomDietrich: *fetched

extensions/documents/homie_last_updated_extension.md
```diff
@@ -0,0 +1,73 @@
...
+
+**Client considerations**
+
+Note that MQTT is not atomic, and does not guarantee any order. This means that even though both the value and the last-update timestamp are published together and in the oder above, the clients may receive these messages in any order, or one of them may be lost. Therefore, **$last-updated** is entirely **best-effort**. It **must** **not** be used for any critical operation.
```
ThomDietrich: comma after atomic?

Because of your argumentation I would actually not state any publishing order up top. If reception in a specific order can't be promised the controller can't act on it either.  Thinking about this some more I now understand one potential purpose of the 0 value. We could define in the extension that a controller has to expect an updated timestamp with every property value update as long as the $last-updated topic is set.

ThomDietrich (changes requested):
Hey!

Some thoughts on your way the two topics are synchronized and what the value 0 stands for. Imho it would be good to give clearer instructions how to process data for the discovering client/controller.

Like the extension 👌 Best!

maxberger:
Sorry for the late reply, my life kept me busy. I am now back and trying to get this into good shape.

I do see a strong push for consistency from multiple side, and I would like to propose yet another alternative. 

Option 1: Keep the proposal as-is, addressing the comments above.

Option 2: Completely re-think the prosposal and provide both the value and the last-updated in the same property. So here is what we could do:
* The extension would be a more generic "value-with-metadata" extension.
* The original value would always be published as normal, but
* An additional "extended" value would be written. This value would follow a common format (probably JSON), and would contain additional meta information, which combines both the value AND the last-updated, and possible other things in the same update

Example:
* homie/super-car/wheels/pressure/ → 32
* homie/super-car/wheels/pressure/$ext → "{value: 32, last-updated: 123456}"

Of course, here we could run into the same consistency issues, but at least a client could look at either one of the two values. It would also eliminate the need for a "0" timestamp.

Pros:
* Extensible
* In it self consistent
* Writing JSON is cheap.

Cons:
* Duplication of values
* clients which would need the extended data would need to parse JSON

wdyt?

davidgraeff:
> Writing JSON is cheap.

Homie has deliberately not used JSON anywhere. For extensions this should be fine, especially if multiple values need to be kept in sync. It if is a fixed known amount of values, CSV could be an option as well though.

ThomDietrich:
Hey all, 
I see how option 1 is flawed (https://github.com/homieiot/convention/issues/191#issuecomment-580298368) but option 2 I would avoid as well.First of all as @davidgraeff said, we decided to stay clear of JSON and we should try to uphold this principle for consistency.
Furthermore, an extension for "value with metadata" feels too far fetched. Remember that one can always implement additional topics with arbitrary complexity in parallel to what is defined by Homie. The Homie convention is here to aid automatic discovery of devices. An extension to process a last-updated makes sense, some generic metadata does not so much.

Following my interpretation of use cases for this extension (@maxberger this should be part of the intro) I propose a **third option**: The extension defines two topics:
* `$last-updated` - epoch time as defined so far, e.g. "1579965965"
* `$last-value-updated` - A combination of value and epoch time, e.g. "12.5;1579965965"

This combination offers a dateime-only topic to directly get a timestamp for keep-alive use cases, and an atomic value-datetime pair for instrumentation/monitoring use cases. Justification for having both would be the ease of processing on both the client and controller side.

Does this cover all use cases? Does anyone dislike the redundancy?

extensions/documents/homie_last_updated_extension.md
maxberger: changed id again

extensions/documents/homie_last_updated_extension.md
maxberger: Rewritten to make it clearer

extensions/documents/homie_last_updated_extension.md
maxberger: Added example, renamed to controller, added (hopefully) better explanation for the 0 value.

extensions/documents/homie_last_updated_extension.md
maxberger: If they are, I prefer to be overly-specific in this case.

extensions/documents/homie_last_updated_extension.md
maxberger: It is hard to "unset" a topic in MQTT; we could publish blank or " ", but that would then not be a valid integer. Using 0 has the advantage that controller which don't want to interpret the value specially will have a valid, transformable timestamp. 

Added this to the doc.

extensions/documents/homie_last_updated_extension.md
maxberger: fixed.

extensions/documents/homie_last_updated_extension.md
maxberger: changed to*should* above.

maxberger:
Like the idea; added this to the extension as well.

extensions/documents/homie_last_updated_extension.md
```diff
@@ -0,0 +1,134 @@
...
+
+Version: **<!--VERSION-->1.0.0<!--VERSION-->**
+Date: **<!--DATE-->17. Feb 2020<!--DATE-->**
+Authors: **<!--AUTHORS-->Max Berger<!--AUTHORS-->**
+License: **<!--LICENSE-->[CCA 4.0](https://homieiot.github.io/license)<!--LICENSE-->**
+
+## Abstract
+This extension adds best-effort last-updated timestamps to properties.
+
+Homie devices will publish a last-updated timestamp along with any property value changes. The last-update timestamp can be sent by devices or gateways that do not constantly pull data from the real back-end. In this case, they can report when they last actually updated the data, so that controllers can act accordingly, and maybe ignore old data, or display this information in an UI.
```
ThomDietrich, line 11: General style thing: Please break the whole article down to one sentence per line. The rendered output will be the same but this makes a few things easier.

ThomDietrich, line 11: I would suggest to start with a rationale sentence or even section, something like

> In most Homie application scenarios it is implicitly assumed that a property update on MQTT directly follows the real world event, e.g., the press of a physical switch would trigger an MQTT message and a controller would react to said message and turn a light on.
>
> This extension addresses the special use case of a Homie device in which the timeliness of an event can't be guaranteed or in which the exact timestamp of a value is of importance.
> Examples for such use cases are
> * ...

ThomDietrich: You sometimes write "last-update", sometimes "last-updated". I understand the incentive but this might be confusing.

extensions/documents/homie_last_updated_extension.md
```diff
@@ -0,0 +1,134 @@
...
+
+A common use case for this is a gateway software which publishes device data in Homie format, but actually uses external sensors to get the actual data. Gateways may re-publish the values the gather from the backend, even if these values are older. Using the last-update timestamp, a controller is then able to distinguish between actual value updates, and value updates which are just a republication of older values.
```
ThomDietrich, line 13: 👍

extensions/documents/homie_last_updated_extension.md
```diff
@@ -0,0 +1,134 @@
...
+
+There are two extensions in this specification. 
+
+The ID of this first extenstion is `org.homie.ext.last-updated`.
+Therefore the **$extensions entry** is `org.homie.ext.last-updated:1.0.0:[3.0.1;4.x]`.
+
+The ID of this extension is `org.homie.ext.last-value-updated`.
+Therefore the **$extensions entry** is `org.homie.ext.last-value-updated:1.0.0:[3.0.1;4.x]`.
+
+A device may implement zero, one, or both extensions. The extension entries refer to the matching property attributes.
```
ThomDietrich, line 28: I suggest to add:

> A controller supporting the extension must support both topics

ThomDietrich: I would not separate the extension into two.
It feels rather confusing and both topics are for the same purpose behind this single extension...

extensions/documents/homie_last_updated_extension.md
ThomDietrich: Unsetting is technically possible. See "Delete": https://www.hivemq.com/blog/mqtt-essentials-part-8-retained-messages/

However, to determine that a Homie device uses the extension publishing any value to the topic (timestamp or 0) is the right way forward.

extensions/documents/homie_last_updated_extension.md
```diff
@@ -0,0 +1,134 @@
...
+This extension defines no nested node attributes.
+
+
+### Property Attributes
+
+This extension defines two new property attributes.
+
+#### $last-updated
+
+The **$last-updated** property attribute is **optional**.
```
ThomDietrich, line 54: "Optional" renders the whole extension ambiguous.

Let's define the following: A Homie device may implement neither of the topics, one topic, or both topics. A Homie controller supporting the extension must support both topics.

With that the rule is simple and deterministic:
A Homie controller supporting the extension shall always evaluate `$last-value-updated` when defined, otherwise `$last-updated` and the property value.

If we agree so far, we can go one step further:
Both topics can be combined into one as the controller is able to parse which of the two formats is used and there is no case in which it actually makes sense for a Homie device to publish both formats.
Wdyt?

ThomDietrich (changes requested):

ThomDietrich:
There is one further design issue we need to discuss. @davidgraeff please add your two cents.

In the Homie convention dollar topics are used to provide metadata for controller-side feature discovery of a device. Your "last-updated" topic communicates data. This is clearly not in line with the convention. Therefore, I think we need to discuss one further change to the topic structure of this extension. Wdyt?

mjcumming:
@maxberger to get around this problem I created a jython script 

https://community.openhab.org/t/asynchronous-events-to-synchronous-event/95198

mjcumming:
@maxberger  and @ThomDietrich 

see http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718105 from @Thalhammer  in [#200](url)  There is ordering of MQTT messages of the same QoS.

Would adding a property attribute $lastupated and specifying that this attribute must be updated after the value has been set solve this is?

Thalhammer:
> @maxberger and @ThomDietrich
> 
> see http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718105 from @Thalhammer in [#200](url) There is ordering of MQTT messages of the same QoS.
> 
> Would adding a property attribute $lastupated and specifying that this attribute must be updated after the value has been set solve this is?

While acording to the rfc MQTT *should* be ordered withing one qos value, I don't think we should rely to badly on it. Not all brokers might follow the spec that closely.

maxberger:
I no longer have the time to follow up with this.

Issue #150, opened 2018-12-24
Retained set commands


MofX:
Is there a reason, why set command are not allowed to be retained?

When I implemented my first device using mqtt, I liked the possibility to retain the color value send from openhab in mqtt, because my device was able to regain the value after it lost power or connection.

The only homie compliant way to achieve  this is using a rule in openhab to detect the device coming back online. Or maybe subscribe to the retained property (without /set), to get the last value.

ThomDietrich:
That indeed seems like a not yet covered use case for Homie. Thanks for bringing up this logical issue.

So far the set command is intended as a one-time event to instruct the device. A set command does not (necessarily) define the actual state of the device at a given point, nor should it be valid and wrongfully (?) be re-evaluated at a later point. Because of that reasoning the set topic is non-retained.

The only valid use of a re-evaluated (retained?) set command is the case of "Restore state after disconnect/power-off" as you described it. I don't like the idea to abuse the set topic by sending retained messages though.

We should discuss alternative solutions.

jamesmyatt:
Remember that retained messages are only sent by the broker in response to a new subscription. So if Homie doesn't reconnect with a clean session, then the retained messages won't be resent. If Homie does connect with a clean session then any QoS 1 and 2 messages waiting for it will be discarded.

davidgraeff:
The property values are retained (or allowed to be retained). A homie device could after initialization restore its internal states by those retained values (if it wants too). Is that not enough?

ThomDietrich:
@davidgraeff I believe everyone agrees there. If I am correct, then @MofX's intend was to discuss the use case and define a solution, e.g. in the implementation guideline for clients or in the FAQ. Correct?

riemp:
@davidgraeff  How can I retain the property values?

homie/devicename/nodename/property/$retained  -->true 

Is this the way to do it?
It isn't working for me.

davidgraeff:
@riemp If it works depends on your implementation and not on the Homie convention. You are in the wrong repository.

The convention states that if $retained is set to true, then the implementation is required to retain the mqtt topic value.

riemp:
@davidgraeff 
I'm using homie convention in openhab(2.4)autodiscovery

davidgraeff:
I'm not taking about the controller. The controller is using the /set topic. I'm talking about your client IoT device esp8266 or whatever.

As you can see in this topic, we agreed that /set topics will never be retained.

riemp:
@davidgraeff
[...]

Issue #148, opened 2018-12-03
Handling device tree changes


andig:
The homie protocol allows updates to the device tree, including unpublishing devices and/or adding and removing nodes and properties. I'm wondering how to best handle this on client side without generating excess traffic?

My thought was to listen for `homie/#` to get updated on device tree changes but that will get hit by all value updates as well. I assume `homie/+/+` should work for devices + nodes (and the properties defined on the node).

Does that make sense?

ThomDietrich:
For me your question would make more sense if you better explained your use case. Constant changes to the device tree should not be normal but these are of course supported. Are you sure that is needed in your use case? Your controller (i.e. the other device observing and controlling a homie device - is that what you called client above?) would need to actively re-discover your homie devices and nodes. This could e.g. be `homie/+/$nodes` or `homie/+/$state` or `homie/+/+/$properties`.

@davidgraeff is any type of automatic re-discovery implemented in the openHAB binding?

davidgraeff:
The openHAB controller subscribes to every found node and property separately and reacts to $properties and $nodes changes.

andig:
> For me your question would make more sense if you better explained your use case.

My use case is a homie to [volkszaehler](https://volkszaehler.org) bridge. As long-running daemon process and data concentrator it must be able to handle device tree changes.

> would need to actively re-discover your homie devices and nodes. This could e.g. be `homie/+/$nodes` or ... `homie/+/+/$properties`.

Perfect.

An (edge) case would be updates to existing properties. The spec does not say so but imho- to limit complexity- it should not be allowed to update existing properties in their abilities, e.g. change `$dataype`. Instead, a property should be unpublished and republished including `$properties`. If that makes sense it should be added to the spec?

davidgraeff:
If the original Homie creators would have wanted a 100% fixed property after publishing, they would have chosen a single topic for property description (eg $config), like the HomeAssistant MQTT component definition.

On the other hand, the openHAB controller is not able to react to $datatype or $format changes without property republishing.

A concise sentence somewhere in the Property section to state that it is undefined behaviour if any of those attributes are changed might be helpful, indeed.

andig:
> On the other hand, the openHAB controller is not able to react to $datatype or $format changes without property republishing.

Neither would I want to add this complexity.

> A concise sentence somewhere in the Property section to state that it is undefined behaviour if any of those attributes are changed might be helpful, indeed.

How about:

[...]

Issue #166, opened 2019-06-09
QoS for non-retained properties


ghost:
Hi!

Would it make sense to specify QoS 2 for non-retained properties, or mention it as a option/reminder for those?
For event based things (doorbell, ish, think more pushbutton) the amount of events might matter. In which case QoS 2 becomes significant.

kollokollo:
Yes, that would make sense. Also think of a GUI, where the user presses a button, to trigger a specific action on a remote device. This should be triggered exactly once.

piegamesde:
Let's continue the discussion from the PR review thread here, so it won't be hidden once the topic is resolved.

@Thalhammer wrote:

> I get why the set command should be QoS 2 in some cases, but why the value ?
> Controllers should be able to handle the same value arriving twice just fine.
> The problem with mixing QoS values is that mqtt does not guarantee ordering for different QoS values, only for the same QoS value.
> See [here](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718105).
> So if we mix QoS 1 and 2 for values sent by the device the $state ready message might arrive before the value of a property.
> 

@mjcumming:

> @Thalhammer thank you that link explaining message ordering. My prior understanding was there no guarantee on ordering.
> 
> The use case for a non retained properties is not clear to me.
> 
> For instance: door bell button, a device to controller event. This is a time sensitive event. If the controller/broker is offline, then there is really no need for the controller to get that message after more than a minute or so.
> 
> Or
> 
> For instance: a single command (brew coffee), a controller to device event. This is a time sensitive command. If the coffee maker doesn't get the command within a minute or so then it should probably never get the command.

piegamesde:
@mjcumming You are making the assumption that all (or most) events are of a time critical nature. This may not be true for everything, but I also can't think of good counter examples.

This leads me to the idea to send non-retained properties with a QoS of 0 (*at most once*).

An alternative solution could be to ignore the fact that messages with QoS 2 may arrive at a later time. Handling that case for time-ciritical events would then be implementation defined, e.g. by passing some time stamps as value for the property.

mjcumming:
Perhaps the logic for dealing with this should reside on the controller. The controller would know if the set command was not acted on/received as the value of the property would not change.

Thalhammer:
How about an attribute similar to $settable to indicate if the controller should use retained or non retained set ?

mjcumming:
[...]

PR #200, opened 2020-04-25
Add implementation details and other things

piegamesde:
I spent two hours reading issues and discussions so I thought I may as well put another two hours into writing down their conclusions (or at least what I think).

Any of the commits may be discussed separately and I'm open to create a distinct pull request for each if there's need for it.

This addresses topics from #150, #51, #148 and probably a few others.

### DIFF
--- a/convention.md
+++ b/convention.md
@@ -211,6 +211,7 @@ Each property must have a unique property ID on a per-node basis which adhere to
 
 * Properties can be **retained**.
   A property is retained by default. A non-retained property would be useful for momentary events (door bell pressed).
+  Non-retained properties must be sent with a QoS of **Exactly once (2)** (this applies for both the value and the `/set`-command, but not for any of the attributes).
 
 A combination of those flags compiles into this list:
 
--- a/convention.md
+++ b/convention.md
@@ -152,7 +152,8 @@ The `$state` device attribute represents the current state of the device.
 There are 6 different states:
 
 * **`init`**: this is the state the device is in when it is connected to the MQTT broker, but has not yet sent all Homie messages and is not yet ready to operate.
-This state is optional, and may be sent if the device takes a long time to initialize, but wishes to announce to consumers that it is coming online. 
+This state is optional, and may be sent if the device takes a long time to initialize, but wishes to announce to consumers that it is coming online.
+A device may fall back into this state to do some reconfiguration.
 * **`ready`**: this is the state the device is in when it is connected to the MQTT broker, has sent all Homie messages and is ready to operate. This implies that the value of every retained property must have been published during initialization. A Homie Controller can assume default values for all optional topics.
 * **`disconnected`**: this is the state the device is in when it is cleanly disconnected from the MQTT broker.
 You must send this message before cleanly disconnecting.
@@ -162,6 +163,12 @@ You have to send this message before sleeping.
 You must define this message as LWT.
 * **`alert`**: this is the state the device is when connected to the MQTT broker, but something wrong is happening. E.g. a sensor is not providing data and needs human intervention.
 You have to send this message when something is wrong.
+
+The following MQTT topics must remain unchanged when a device is in `ready`, `sleeping` or `alert` state:
+
+* Any device attributes except `$name` and `$state`
+* The `$properties` attribute of any node
+* Any attribute of any property except `$name`
 
 ### Nodes
 
--- a/convention.md
+++ b/convention.md
@@ -153,7 +153,7 @@ There are 6 different states:
 
 * **`init`**: this is the state the device is in when it is connected to the MQTT broker, but has not yet sent all Homie messages and is not yet ready to operate.
 This state is optional, and may be sent if the device takes a long time to initialize, but wishes to announce to consumers that it is coming online. 
-* **`ready`**: this is the state the device is in when it is connected to the MQTT broker, has sent all Homie messages and is ready to operate. A Homie Controller can assume default values for all optional topics.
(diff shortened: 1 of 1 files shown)


mjcumming (commented):
For device reconfiguration, I am not sure there is a requirement for the device to go offline or to change its state to init? @Thalhammer ?

piegamesde:
Thanks for correcting the typo. Currently, the convention doesn't say anything about any changes or reconfiguration. I just thought it would make writing Homie-Controllers easier if they didn't have to be aware of any device changes all of the time (It may also make writing libraries easier). I remember some discussion about adding an explicit `reconfiguring` state, but I think a simple reconnection or using the `init` state should be enough.

mjcumming:
I think that makes sense. I'll let @Thalhammer weigh in on the proposal before approving.

convention.md
```diff
@@ -204,6 +211,7 @@ Each property must have a unique property ID on a per-node basis which adhere to
 
 * Properties can be **retained**.
   A property is retained by default. A non-retained property would be useful for momentary events (door bell pressed).
+  Non-retained properties must be sent with a QoS of **Exactly once (2)** (this applies for both the value and the `/set`-command, but not for any of the attributes).
```
Thalhammer: Do we really want that ?
QOS 2 is something thats really hard to support on devices without persistent storage.
Shouldn't QOS 1 be enough ?

convention.md
```diff
@@ -324,3 +332,28 @@ For example, an organization `example.org` wanting to add a feature `our-feature
...
+Some devices require knowledge of their settable retained properties to function properly.
+The homie convention does not specify how to initialize/recover them e.g. after a power cycle.
+A few common approaches are:
+
+* A device can simply load default values from some configuration file.
+* A device can restore its previous state from some local storage. This is the recommended way.
+* A device may try to restore its state using MQTT. This can be done by subcribing to the respective channels.
+An alternative way is to recover the state from other MQTT channels that are external to the Homie specification.
+This is not a recommended approach though, because retained messages are only sent by the broker in response to a new subscription.
+So if a device doesn't reconnect with a clean session, then the retained messages won't be resent.
```
Thalhammer: If a device connects with a non clean session it is assumed to "know" what happend in the last connection (mqtt kind of defines this). This means it should still know the retained values sent. If there is no persistence it should really be connecting with clean session set.

convention.md
piegamesde: See #166 for the discussion about this. I don't know really much about the MQTT implementation internals and their implications.

Would it be a solution to say that non-retained properties ***should*** be sent with QoS 2? So all devices that are capable will do it and the few that can't don't have to.

convention.md
piegamesde: I'm sorry, but I don't understand. What exactly is a "clean session"? What exactly is wrong about the paragraph and what should a more correct version?

My motivation behind this was to say that this approach at restoring state is a bad idea, and to explain why.

Thalhammer: MQTT supports keeping sessions while the device is offline.
If you connect and request a clean session you will get one that looks as if you never connected before.
However if you dont request a clean session the broker will restore all subscriptions you had the last you where connected.  It will also deliver all messages sent to a subscription with qos 1 or higher while the device was offline.

You would normally connect with a clean session if your device state is lost (like after a reboot without persistent storage, or after a reset) and connect to an existing session otherwise if you're device still knows what state its world is in (like a connection drop/reconnect).

The only difference is that with an existing session you dont get retained topics unless you unsubscribe and resubscribe them again or they changed while the device was offline. Also there is no need to subscribe if you have an existing session as the broker keeps subscriptions alive.

convention.md
Thalhammer: I get why the set command should be QoS 2 in some cases, but why the value ?
Controllers should be able to handle the same value arriving twice just fine.
The problem with mixing QoS values is that mqtt does not guarantee ordering for different QoS values, only for the same QoS value.
See [here](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718105).
So if we mix QoS 1 and 2 for values sent by the device the $state ready message might arrive before the value of a property.

mjcumming: @Thalhammer thank you that link explaining message ordering. My prior understanding was there no guarantee on ordering.

The use case for a non retained properties is not clear to me. 

For instance: door bell button, a device to controller event. This is a time sensitive event. If the controller/broker is offline, then there is really no need for the controller to get that message after more than a minute or so. 

Or

For instance: a single command (brew coffee), a controller to device event. This is a time sensitive command. If the coffee maker doesn't get the command within a minute or so then it should probably never get the command.

convention.md
mjcumming: @homieiot/involved in my implementations, the Homie devices know their state ie on/off, dimmer percentage, temperature etc. Is there a situation where a device would not know it's own state and need to restore a value?

As part of other discussion regarding QoS, the Homie spec probably needs to address controller set messages that are not delivered in a timely fashion. Part of our discussion in [#150](url)

piegamesde: @mjcumming The section I wrote about this is limited to settable properties that can't easily be recovered. Think of music players for example. Or in my case, a light effects controller.

mjcumming (commented):
I think this verbiage is clarifies the ready state.

convention.md
piegamesde: I've removed that sentence since the topic is still subject to discussions.

mjcumming (approved):

convention.md
Tieske: didn't want to revive an old thread, so started a new issue wrt this discussion: https://github.com/homieiot/convention/issues/247

PR #224, opened 2021-11-16
clarify $format being required for enum and color

Tieske:
Both `enum` and `color` cannot be set if the `$format` topic is not specified. This can be deducted from the spec, but this clarifies it explicitly.


schaze (approved):

Tieske:
@schaze would you mind having a look at https://github.com/Tieske/convention/pull/2 (same change, targeted at v5)

PR #239, opened 2023-01-15
fix(property) handling empty-string values

### LINKED ISSUE #223: Clarify handling of empty strings
There is an ambiguity in the specification regarding empty strings.

Properties with a string value might be empty, the specification states that "An empty string ("") is a valid payload".

Also section 7.2 specifies that "Devices can remove old properties and nodes by publishing a zero-length payload on the respective topics." which poses the question how a message deleting a property differs from a message setting a string-property to the empty string.

The specification needs to specify what bytes to set as payload to represent the empty string. One option might be to set the payload to a single "zero" byte to represent an empty string.

Tieske:
fixes #223


schaze (approved):

Tieske:
I do have second thoughts about this one. Since 0x00 is a terminator for strings in many applications/languages, it might be safer to use another value like 0x01. Better safe than sorry...

Issue #220, opened 2021-04-08
$state and $homie can be empty


daenney:
Based on the spec, it appears that both `$state` and `$homie` can be empty:

https://homieiot.github.io/specification/#device-attributes:
> The following device attributes are mandatory and MUST be send, even if it is just an empty string.
> Topic | Description
> -- | --
> $homie | The implemented Homie convention version
> $name | Friendly name of the device
> $state | See Device Lifecycle
> $nodes | Nodes the device exposes, separated by , for multiple ones.
> $extensions | Supported extensions, separated by , for multiple ones

If `$homie` is empty then I don't know which version of the convention the device implements, which complicates supporting different versions in the same network. I could make the assumption that empty means `4.0.0` but that seems wrong too. It feels like `$homie` should never be allowed to be empty. Even when a device is still in its initialising phase, it should already know which version of the convention it's going to publish itself as. The alternative would be to interpret a device with an empty `$homie` as the signal for "please forget about me" which would cause the controller to fully remove the device?

`$state` being empty also seems strange. The device lifecycle specifies a number of states a device can be in, and none of them is empty. It seems more practical that empty shouldn't be allowed and a device should default to `init` if it hasn't explicitly published a different state?

Tieske:
good point. This also touches on #223 since empty strings mean deleting the topic in most MQTT servers.

PR #242, opened 2023-01-16
fix(device) be more specific about empty attributes 

Tieske:
companion for #241, see also #220
on proposed v5 of the spec


Tieske:
merging this based on approval for #241

PR #243, opened 2023-01-16
fix(percent) drop the percent type 

### LINKED ISSUE #232: drop the Percent type
The spec is ambiguous about this type, since the type is listed, but then it mentions it is a float or integer with `unit` being set to `%`. So should an implementation accept a Percent type otronly floats and integers?

Drop the `Percent` type from paragraph `2.2 Payload` would clear this up.

Tieske:
fixes #232


schaze (approved):
Never made any sense anyways to mention this specifically as the specification covers this in general with the unit field.

PR #245, opened 2023-01-17
readability(retained) reorder more logically

Tieske:
has no functional changes


Tieske:
just minor reorder, merging this.

PR #249, opened 2023-01-18
feat(node) add a set topic to set multiple values at once 

Tieske:
On a node level, set multiple properties at once.


Tieske:
discussed to not target this one at Homie5 for now. Can be added later.

devbotas:
I'm not a fan of adding a JSON to the party. Homie topics are already structured, so adding another layer of structure feels unnecessary.

Tieske:
The problem at hand: setting multiple properties at once, to prevent inconsistent states due to MQTT message delivery ordering (which one cannot rely on).

This proposal uses JSON to wrap multiple values into a single value. The concept; wrapping multiple values into one, will always be necessary to solve the problem. So imho this will always require some sort of encoding.

PR #255, opened 2023-03-05
feat(datatype) add JSON datatype. Extend color type with xyz color space

schaze:
As initially discussed v5 needs to support complex datatypes. This PR adds a JSON datatype for this.
There is also a color datatype extension with a xyz color space format included which is widely used in different smartbulbs (e.g. Philips Hue)

### DIFF
--- a/convention.md
+++ b/convention.md
@@ -88,11 +88,12 @@ needs this, then it should provide for an escape mechanism on application level.
  
 #### Color
 
-- Color payload validity varies depending on the property format definition of either "rgb" or "hsv"
+- Color payload validity varies depending on the property format definition of either "rgb", "hsv" or "xyz"
 - Both payload types contain comma separated numbers of differing restricted ranges. The numbers must conform to the [float](#float) format
 - The encoded string may only contain the [float](#float) numbers and the comma character ",", no other characters are permitted, including spaces (" ")
 - Payloads for type "rgb" contain 3 comma separated values of [floats](#float) with a valid range between 0 and 255 (inclusive). e.g. 100,100,100
 - Payloads for type "hsv" contain 3 comma separated values of [floats](#float). The first number has a range of 0 to 360 (inclusive), the second and third numbers have a range of 0 to 100 (inclusive).  e.g. 300,50,75
+- Payloads for type "xyz" contain 2 comma separated values of [floats](#float) with a valid range between 0 and 1 (inclusive). The z value can be calucated via z=1-x-y and is therefore not transmitted. (see [CIE_1931_color_space](https://en.wikipedia.org/wiki/CIE_1931_color_space))
 - An empty string ("") is not a valid payload
 
 #### DateTime
@@ -110,6 +111,10 @@ H: Indicates hour and is preceded by the number of hours, if hours are specified
 M: Indicates minutes, and is preceded by the number of minutes, if minutes are specified.
 S: Indicates seconds, preceded by the number of seconds, if seconds are specified.
 - An empty string ("") is not a valid payload
+
+#### JSON
+- Contains a JSON string for transporting complex data formats that cannot be exposed as single value attributes
+- For validation of the JSON data the `format` attribute may contain a [JSONSchema](http://json-schema.org/) defintion as string.Recommended version is Draft-07 http://json-schema.org/draft-07/schema# as this is currently most widely supported by most tools and libraries. This is optional but strongly recommended.
 
 ## Base Topic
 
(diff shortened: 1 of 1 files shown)


convention.md
```diff
@@ -88,11 +88,12 @@ needs this, then it should provide for an escape mechanism on application level.
  
 #### Color
 
-- Color payload validity varies depending on the property format definition of either "rgb" or "hsv"
+- Color payload validity varies depending on the property format definition of either "rgb", "hsv" or "xyz"
 - Both payload types contain comma separated numbers of differing restricted ranges. The numbers must conform to the [float](#float) format
 - The encoded string may only contain the [float](#float) numbers and the comma character ",", no other characters are permitted, including spaces (" ")
 - Payloads for type "rgb" contain 3 comma separated values of [floats](#float) with a valid range between 0 and 255 (inclusive). e.g. 100,100,100
 - Payloads for type "hsv" contain 3 comma separated values of [floats](#float). The first number has a range of 0 to 360 (inclusive), the second and third numbers have a range of 0 to 100 (inclusive).  e.g. 300,50,75
+- Payloads for type "xyz" contain 2 comma separated values of [floats](#float) with a valid range between 0 and 1 (inclusive). The z value can be calucated via z=1-x-y and is therefore not transmitted. (see [CIE_1931_color_space](https://en.wikipedia.org/wiki/CIE_1931_color_space))
```
Tieske, line 96: Is it wise to have 3 color systems? This forces every controller to support all 3 of them, and do all conversions. Wouldn't it make more sense to support 1 only, and have the conversion logic only where it is needed?

convention.md
```diff
@@ -111,6 +112,10 @@ M: Indicates minutes, and is preceded by the number of minutes, if minutes are s
 S: Indicates seconds, preceded by the number of seconds, if seconds are specified.
 - An empty string ("") is not a valid payload
 
+#### JSON
+- Contains a JSON string for transporting complex data formats that cannot be exposed as single value attributes
```
Tieske, line 116:
```suggestion
- Contains a JSON string for transporting complex data formats that cannot be exposed as single value attributes.
- The value MUST be either a JSON-Array or JSON-Object type, for other types the standard Homie types should be used.
```

convention.md
```diff
@@ -337,7 +342,7 @@ The format attribute specifies restrictions or options for the given data type.
...
 | integer      | no       | `:`      | `min:max` where min and max are the respective minimum and maximum (inclusive) allowed values, both represented in the format for [integer types](#integer). Eg. `5:35`. If the minimum and/or maximum are missing from the format, then they are open-ended, so `:10` allows a value <= 10 |
 | enum         | yes      |          | A comma-separated list of non-quoted values. Eg. `value1,value2,value3`. If quotes or whitespace are used adjacent of the '`,`' (comma), then they are part of the value. Individual values can not be an empty string, hence at least 1 value must be specified in the format. |
-| color        | yes      |          | `rgb` or `hsv`. See the [color type](#integer) for the resulting value formats. |
+| color        | yes      |          | `rgb`, `hsv` or `xyz`. See the [color type](#color) for the resulting value formats. |
```
Tieske, line 345: This table should also get an entry for `json`

convention.md
```diff
@@ -111,6 +112,10 @@ M: Indicates minutes, and is preceded by the number of minutes, if minutes are s
 S: Indicates seconds, preceded by the number of seconds, if seconds are specified.
 - An empty string ("") is not a valid payload
 
+#### JSON
+- Contains a JSON string for transporting complex data formats that cannot be exposed as single value attributes
+- For validation of the JSON data the `format` attribute may contain a [JSONSchema](http://json-schema.org/) defintion as string. Recommended version is Draft-07 http://json-schema.org/draft-07/schema# as this is currently most widely supported by most tools and libraries. This is optional but strongly recommended.
```
Tieske, line 117: This line should probably go in the "format" section.

convention.md
Tieske: Some comments;
 - since this format field goes in the JSON description document, the schema itself does not need to be a sytring, but can be a json object (otherwise you'd get escaped json in a string).
 - if there is no schema provided, then the minimum validation should be that the data is a valid array or object.
 - I think we should refrain from recommending json-schema versions, but just list the challenges and caveats of using them, so folks are warned to do the right-thing™️

convention.md
Tieske: that could mean device implementers doing this, so in hind sight I think doing it on the controller side is fine.

convention.md
Tieske: as for validation, the default `format` for a `json` type should be;
```json
{"anyOf": [{"type": "array"},{"type": "object"}]}
```

schaze:
Replaced with #273 and #274

PR #261, opened 2023-06-11
chore(id): simplify ID checks

Tieske:
by removing the constraint that an ID cannot start nor end with a '-', the validation becomes a lot simpler. Eg. in Lua which has a pattern matching engine, but not full regex capabilities, the current ID cannot be validated with a single pattern.

A regex would now simply be `^[a-z0-9-]+$` (or `^$*[a-z0-9-]+$` to match homie-keys), also way easier to implement without a regex engine.  Alternatively we could require the ID to be minimum 2 characters in length, which would result in `^[a-z0-9][a-z0-9-]*[a-z0-9]$`, which is still fairly simple.

The current limitation is just for human readability, so even if we remove it, it is most unlikely that one would use that anyway. And as mentioned, it is NOT a technical limitation.


schaze (approved):
Good suggestion. I agree that the hyphen part has no technical significance.

PR #262, opened 2023-06-12
feat(log): replace "alert" state with logging

Tieske:
The "alert" state currently is more of an event than an actual state. Hence it should be removed and replaced with a notification. Since the common way of notifying is logging, I created a description on how to do this by introducing an optional device topic `$log` which can be used in a generic way.


schaze:
I would keep the alert state and only add the $log part in your commit.
The alert state is a simple and straight forward way to identify devices that need attention or may not be used as normally intended. With the proposed solution I would need to check 5 different topics for a potential problem of which only 3-4 would indicate a most likely actionable "alert".

schaze (approved):

PR #263, opened 2023-06-12
feat(property): add an optional $target property

Tieske:
in case a property change is not instantaneous

### DIFF
--- a/convention.md
+++ b/convention.md
@@ -36,10 +36,11 @@ root device (the device at the root of the parent-child tree).
 MQTT will treat an empty string payload as a "delete" instruction for the topic, therefor an
 empty string value is represented by a 1-character string containing a single byte value 0 (Hex: `0x00`, Dec: `0`).
 
-The empty string (passed as an MQTT payload) can only occur in 2 places;
+The empty string (passed as an MQTT payload) can only occur in 3 places;
 
 - `homie5` / `device ID` / `node ID` / `property ID`; reported property values (for string types)
 - `homie5` / `device ID` / `node ID` / `property ID` / `set`; the topic to set properties (of string types)
+- `homie5` / `device ID` / `node ID` / `property ID` / `$target`; the target property value (for string types)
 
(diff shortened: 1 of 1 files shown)


schaze:
I know we already talked about this and I also understand the physical behavior this wants to address. But thinking about it now I really fail to find any actual use-case where the $target property would be of any significance. 
With the current and the proposed spec a device has three options for their state update in the brightness example in the commit:

1. emit 100 as new value directly even though the 100 brightness will only be achieved 5 seconds later
2. emit any intermediate value state every second (or any other arbitrary fixed or random interval step) until 100 is reached
3. emit nothing for 5 seconds and then update the value to 100

For every option a $target property could be optionally published as a first step.
But if we would publish a $target property what could any other device or controller do with this information? What scenario would there be where I would need the $target value field?

Tieske:
It's about 2 things;
- closing the control loop for a controller (it knows it's command was received) once the `$target` is updated to the requested value.
- user interaction/feedback, by showing both the target and actual, the user knows the system is "working on it"

Some examples for the latter (the light is probably not the best one);
- a valve for underfloor heating that takes appr. 2 minutes to open fully.
  - user clicks "open"
  - devices accepts command, but the valve is not considered "open" until at least 70% open (eg. after 90 seconds appr). If you start the boiler/pump before that it will go into an error state.
  - for 90 seconds the state remains "closed", user is now confused, and repeatedly ticks the "open" button.

Same thing for a screen/blind, change is not instantaneous.

Very typical example; Spotify app, controlling a Sonos speaker
- user changes volume from 20% to 70%; dragging the green dot across the slider from 20 -> 70
- once the user let's go, the dot immediately jumps back to 20
- over the next few seconds with 1 or more intermediate steps the dot jumps up to 70
(I see my kids struggling with this thing every-single-time, they keep dragging the dot again and again, which actually makes it worse)

schaze (approved):

PR #268, opened 2023-06-16
clarify(enum): clarify that whitesapce is significant

Tieske:
previously it mentioned that whitespace should be removed, but it was not clear whether the sender or receiver was supposed to do this. By making whitespace significant, the whole thing becomes easier to understand and to validate.


schaze (approved):
I fully agree. Whitespaces matter and are just as significant as any other character.

PR #270, opened 2023-06-18
change(description): change arrays to objects

Tieske:
by changing the node array (in device) and the property array (in node) to objects, the ID can be removed from the node/property, and it ensures that there never can be duplicates (which could happen with an array)


schaze (approved):
Thanks!

PR #272, opened 2023-06-19
chore(state): add some more details around state management

Tieske:
especially around safely adding/removing child relationships

### DIFF
--- a/convention.md
+++ b/convention.md
@@ -497,26 +497,28 @@ If a device wishes to modify any of its nodes or properties, it can
 Devices can remove old properties and nodes by deleting the respective MQTT topics by publishing an empty message
 to those topics (an actual empty string on MQTT level, so NOT the escaped `0x00` byte, see also [empty string values](#empty-string-values)).
 
+When adding many child devices, implementations should take care of not publishing too many parent-updates, since every controller would have to parse the description again and again.
+
 The recommended way to add/remove child device is as follows (note: due to MQTT message ordering consistency at any stage in this process cannot be guaranteed):
 
 #### Adding:
 
-1. first publish the child-device(s), as any other device
+1. first publish any child-devices, as any other device
     1. set child-device state to `"init"`
     1. publish child-device details (including parent details in `root` and `parent` fields)
     1. set child-device state to `"ready"`
 1. update the parent device, as any other change
     1. set parent state to `"init"`
-    1. update parent description (add the child ID to its `children` array)
+    1. update parent description (add any child IDs to its `children` array)
     1. set parent state to `"ready"`
 
 #### Removing:
 
 1. update the parent device
     1. set parent state to `"init"`
-    1. update parent description (remove the child ID from its `children` array)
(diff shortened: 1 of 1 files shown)


schaze:
I do not like the interweaved coupling of parent and child updates. This will make an implementation very cumbersome. Also you cannot guarantee another mqtt client will receive updates for different topics in sequential manor. In addition I want to keep coupling between parent and child as minimal as possible. Exact sequence and timing in MQTT do not work well in my experience.

I would clearly split this up and also switch the order when adding a child to first make sure the devices are published (with at least $state:init) before adding them to the parent (this way one could add multiple child devices with a single parent update).

Adding: 
* Publish 1..n child(-ren)  (as defined in spec - $state:init, $description, ..., ..., $state:ready)
* update parent's children property (as defined in the spec, $state:init, $description, state:ready)

Removing, whilst leaving the device on the Homie network:
* update parent's children property [remove 1..n children]  (as defined in the spec, $state:init, $description, $state:ready)
* update child-device's description with cleared `"root"` and `"parent"` attributes (as defined in spec, $state: init, $description)
* set child's state to disconnected

Removing, permanently
* update parent's children property [remove 1..n children]  (as defined in the spec, $state:init, $description, $state:ready)
* set child's $state to init
* clear child-device description (clear the `$description` attribute topic)
* clear all child-device topics (order is irrelevant as without a $description there is no device)

Tieske:
@schaze thx, make sense. Updated.

I removed the "_Removing, whilst leaving the device on the Homie network_" part, since that is something that works in theory, but really do not see a real use-case for it.

convention.md
```diff
@@ -486,7 +494,29 @@ If a device wishes to modify any of its nodes or properties, it can
...
+    1. update parent description (add the child ID to its `children` array)
+    1. set parent state to `"ready"`
+
+#### Removing:
+
+1. update the parent device
+    1. set parent state to `"init"`
+    1. update parent description (remove the child ID from its `children` array)
+    1. set parent state to `"ready"`
+1. clear child-device(s) topics, starting with the `$state` topic
```
```suggestion
When adding many child devices, implementations should take care of not publishing too many parent-updates, since every controller would have to parse the description again and again.

The recommended way to add/remove child device is as follows (note: due to MQTT message ordering consistency at any stage in this process cannot be guaranteed):

#### Adding:

1. first publish any child-devices, as any other device
    1. set child-device state to `"init"`
    1. publish child-device details (including parent details in `root` and `parent` fields)
    1. set child-device state to `"ready"`
1. update the parent device, as any other change
    1. set parent state to `"init"`
    1. update parent description (add any child IDs to its `children` array)
    1. set parent state to `"ready"`

#### Removing:

1. update the parent device
    1. set parent state to `"init"`
    1. update parent description (remove any child IDs from its `children` array)
    1. set parent state to `"ready"`
1. clear any child-device(s) topics, starting with the `$state` topic
```

schaze (approved):

PR #274, opened 2023-06-19
feat(format): add color format 'xyz'

Tieske:
replaces #255 the color part

### DIFF
--- a/convention.md
+++ b/convention.md
@@ -90,11 +90,11 @@ needs this, then it should provide an escape mechanism on the application level.
 #### Color
 
 - Color payload validity varies depending on the property format definition of either "rgb", "hsv", or "xyz"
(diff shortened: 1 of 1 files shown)


Tieske:
I added a second commit that will allow a property to support multiple color formats, eg, RGB + XYZ. This will enable homie-library authors to write the color conversion once, and the device author (using the library) will not have to write his own.

For example;
- device supports RGB
- homie lib supports RGB + HSV -> publishes device as supporting RGB+HSV
- controller sends HSV value
- lib convert received HSV to RGB
- device consumes RGB

Obviously this is just a suggestion, it can be removed again.

schaze (approved):
Thanks for splitting up by PR!

Regarding the 2nd commit:
Color conversions are the devil and IMHO should be avoided where possible. In my experience this means whatever the Device can accept natively would be offered as input format. However there are devices that support multiple color formats natively so offering multiple options as input could reduce the amount conversions needed.

PR #276, opened 2023-07-25
fix(log): clarify all messages must be non-retained

Tieske:


schaze (commented):
Looks good.
How about we release it? 5.0.0?

Tieske:
I'm still implementing (takes longer than expected due to time constraints), and occasionally run into issues like this. So not 100% comfortable with releasing just yet. Maybe start a new issue, with a description of what "done" looks like.

PR #278, opened 2023-09-02
Added information about unpublishing/removing a device

schaze:
The specification did not yet contain any information about how a device is to be removed/unpublished from the broker and also how a controller can detect this.
This PR adds this information by specifying a zero length payload has to be sent to the $state topic first before any other topics are cleared.

### DIFF
--- a/convention.md
+++ b/convention.md
@@ -249,7 +249,7 @@ You must define this message as the last will (LWT) for root devices.
 
 In order to unpublish / remove a device the following steps should be performed in order:
 1. remove the retained `$state` attribute from the broker by publishing a zero length payload message to its topic
-2. any other retained attributes or property values can be cleared via the same method afterwards
+2. any other retained attributes or property values should be cleared via the same method afterwards
 
 ### Nodes
 
--- a/convention.md
+++ b/convention.md
@@ -125,8 +125,11 @@ If this root topic does not suit your needs (in case of, e.g., a public broker o
 you can change the first segment, but the `"/5/"` segment must be retained. This allows controllers
 to subscribe to only the devices they are compatible with.
 
+## Auto-Discovery
+
 Homie 5 controllers must by default perform auto-discovery on the wildcard topic `"+/5/+/$state"`.
 Controllers are free to restrict discovery to a specific root topic, configurable by the user.
+A zero length payload published on the $state topic indicates a device removal.
 
 ## Topology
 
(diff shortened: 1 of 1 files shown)


Thalhammer (approved):

convention.md
```diff
@@ -244,6 +247,10 @@ You have to send this message before sleeping.
 You must define this message as the last will (LWT) for root devices.
 * **`alert`**: in this state the device is connected to the MQTT broker, but something is wrong and needs human intervention. The device should be considered inoperable similar to the `init` state. When in this state, it is encouraged to use the [`$log` topic](#logging) to provide details on what is wrong.
 
+In order to unpublish / remove a device the following steps should be performed in order:
+1. remove the retained `$state` attribute from the broker by publishing a zero length payload message to it's topic
+2. any other retained attributes or property values can be cleared via the same method afterwards
```
Tieske: `can` is quite arbitrary, can we make this more specific? maybe "should" or "may" (which are more 'rfc' type speak). I'd prefer "should".

convention.md
```diff
@@ -244,6 +247,10 @@ You have to send this message before sleeping.
 You must define this message as the last will (LWT) for root devices.
 * **`alert`**: in this state the device is connected to the MQTT broker, but something is wrong and needs human intervention. The device should be considered inoperable similar to the `init` state. When in this state, it is encouraged to use the [`$log` topic](#logging) to provide details on what is wrong.
 
+In order to unpublish / remove a device the following steps should be performed in order:
+1. remove the retained `$state` attribute from the broker by publishing a zero length payload message to it's topic
```
```suggestion
1. remove the retained `$state` attribute from the broker by publishing a zero length payload message to its topic
```

Tieske (changes requested):
Should we also add a note on controller behaviour? eg. since they should keep track of revisions of the description JSON by its id.

So should a controller upon detecting a deletion, remove and clear all cached data/knowledge it has on the deleted device?

convention.md
schaze: Done

schaze:
> Should we also add a note on controller behaviour? eg. since they should keep track of revisions of the description JSON by its id.
> 
> So should a controller upon detecting a deletion, remove and clear all cached data/knowledge it has on the deleted device?

I would not specify this in the convention. Any controller is free to handle this any way it likes. We also do not specify what the controller should do with the data of a device when it is discovered. IMHO a device that is not existing any more is of no use so the default would be to clear all data of it, except if there is a specific use case the controller likes to implement for it - whatever that may be.

Tieske:
>  IMHO a device that is not existing any more is of no use so the default would be to clear all data of it, except if there is a specific use case the controller likes to implement for it - whatever that may be.

Agreed, but not specifying it leaves it too open imho, so can we specify this as a "should", as to say; this is expected behaviour, but in rare circumstances one can deviate.

convention.md
```diff
@@ -125,8 +125,11 @@ If this root topic does not suit your needs (in case of, e.g., a public broker o
 you can change the first segment, but the `"/5/"` segment must be retained. This allows controllers
 to subscribe to only the devices they are compatible with.
 
+## Auto-Discovery
+
 Homie 5 controllers must by default perform auto-discovery on the wildcard topic `"+/5/+/$state"`.
 Controllers are free to restrict discovery to a specific root topic, configurable by the user.
+A zero length payload published on the $state topic indicates a device removal.
```
Tieske, line 132:
```suggestion
A zero length payload published on the $state topic indicates a device removal. This indicates the device has no intent to return later, and as such controllers should consider all device related (cached) data to be invalid.
```

Tieske: I've given this some more thought, and I think it is actually a matter of a higher level issue; define when a device exists.

I'll drop this, and create an addition to this PR.

Tieske:
my addition ended up here: https://github.com/schaze/convention/pull/1

Tieske (approved):

PR #289, opened 2024-02-14
Update convention.md

rumbu13:
Added binary data type

### DIFF
--- a/convention.md
+++ b/convention.md
@@ -118,6 +118,12 @@ S: Indicates seconds, preceded by the number of seconds, if seconds are specifie
 - Contains a JSON string for transporting complex data formats that cannot be exposed as single value attributes.
 - The payload MUST be either a JSON-Array or JSON-Object type, for other types the standard Homie types should be used.
 
+#### Binary
+
+- Contains raw bytes of data.
+- Recommended for data that cannot be efficiently represented as string like pictures or audio samples.
+- More information about the data can be provided in the format.
+
 ## Base Topic
 
 The root topic in this document is `"homie/5/"`.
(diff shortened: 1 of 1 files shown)


convention.md
```diff
@@ -364,7 +370,7 @@ the formats for displaying values.
...
 | boolean      | no       | `false,true` | Identical to an enum with 2 entries. The first represents the `false` value and the second is the `true` value. Eg. `close,open` or `off,on`. If provided, then both entries must be specified. **Important**:  the format does NOT specify valid payloads, they are descriptions of the valid payloads `false` and `true`. |
 | json         | no       | `{"anyOf": [{"type": "array"},{"type": "object"}]}` | A [JSONSchema](http://json-schema.org/) definition, which is added as a string (escaped), NOT as a nested json-object. See [JSON considerations](#json-considerations), for some ideas wrt compatibility. If a client fails to parse/compile the JSONschema, then it should ignore the given schema and fall back to the default schema.
-
+| binary       | no       |  `application/octet-stream` | Any media type (also known as MIME type) that indicates the nature and the format of the data. [IANA Media Types](https://www.iana.org/assignments/media-types/media-types.xhtml) is the standardized and official data source for possible values.
```
Tieske, line 373: I don't like this format attribute for the blob. What if the value accepts 2 types; eg. JPG and PNG?

What if those formats cannot be distinguished from the payload itself? Then we'd need to specify the format when setting the values in "the binary blob just set is image/jpg", and we cannot do that.

So I think we should drop the format here. A blob is a blob, without any further rules/policies. The consumer must use proper validations to ensure it is valid.

convention.md
```diff
@@ -118,6 +118,12 @@ S: Indicates seconds, preceded by the number of seconds, if seconds are specifie
 - Contains a JSON string for transporting complex data formats that cannot be exposed as single value attributes.
 - The payload MUST be either a JSON-Array or JSON-Object type, for other types the standard Homie types should be used.
 
+#### Binary
+
+- Contains raw bytes of data.
+- Recommended for data that cannot be efficiently represented as string like pictures or audio samples.
```
Tieske, line 124:
```suggestion
- Recommended for data that cannot be efficiently represented as a UTF-8 string like pictures or audio samples.
```

convention.md
```diff
@@ -118,6 +118,12 @@ S: Indicates seconds, preceded by the number of seconds, if seconds are specifie
...
+- More information about the data can be provided in the format.
+
```
Tieske, line 126: here we need to add that this doesn't have to be valid UTF-8. In other places where we say it has to be UTF-8, the exception should be added.

Tieske (changes requested):

Tieske:
@rumbu13 ping

PR #297, opened 2024-08-03
fix(color): xyz format should include brightness

### LINKED ISSUE #296: Color bug (Homie 5): xyz does not include brightness
In the Homie 5 draft, there is "rgb", "hsv", and "xyz" for the color type.

- in "rgb" the brightness setting is included in the information; rgb(50,50,50) is the same color as rgb(100,100,100), but at half the brightness
- in "hsv" the brightness is independent in the "v" component, where "hs" determines the color
- in "xyz" (specifically if we define `z = 1 - x - y`) only the color is encoded, no brightness.

Problem: 2 of them encode brightness, the third doesn't. _**So they are not equivalent!**_.

In the "xyz" system (CIE 1931) there are actually 2 representations; "XYZ" and "xyz", which are different:

- "XYZ" (Tristimulus Values) encodes the color and the brightness. The values of X, Y, and Z do NOT add up to 1.
- "xyz" (Chromaticity Coordinates) encode only the color, and no brightness. In this case the values of x, y, and z do add up to 1.

The only solution I see to this is to either remove "xyz" as an option, or change the definition of "xyz" (currently Chromaticity Coordinates) to Tristimulus Values, meaning, the last value (z) must also be passed along, and they do not have to add up to 1.

@schaze this originated from you I think. And iirc mostly because some devices (Philips Hue and friends) seem to prefer the "xyz" system. So wdyt?

Tieske:
fixes #296


Tieske:
closing this in favour of #298

PR #300, opened 2024-09-07
chor(node/prop): make 'name' optional

Tieske:
In most cases for nodes and properties the 'id' is the same as the name. So it makes sense to make the name 'optional' to reduce overhead.

The example in the spec:

level | id | name
-|-|-
device | super-car | Supercar
node | engine | Car engine
property | temperature | Engine temperature

In a homie based system, there can be many devices, and id's are sometimes assigned randomly or based on device types (whilst having many of the same types, eg. 10 of the same light bulbs, or relays). So for devices it makes sense to have a mandatory `name` attribute.

But for nodes and properties this isn't the case. Typically the `id` within the context of the device is specific enough. Within the context of device "Supercar", the node "engine" doesn't need the description "Car engine". And similarly the property "temperature" within node "engine" doesn't need the description "engine temperature".

I have some doubts on the device level though; for consistency it might be nice to have all of them be optional...


schaze:
I would suggest to make it optional for all level, also for the device.

Tieske:
updated, also the device name is now optional

schaze (approved):

PR #303, opened 2024-10-15
minor clarification in the base-topic

Tieske:


schaze (approved):
Good catch. I like the distinction. Although I now have to change the name of the constant in my library everywhere ;)

PR #304, opened 2024-10-27
change(version): remove 'higher' constraint

Tieske:
This constraint forces the use of persistent storage. By removing it
we can simply start with a random number, e.g. 0-65535 and then increment
it (with possibly a roll-over). Especially for resource constraint
devices (which may not have storage nor be able to do unix timestamps
which is 32 bits).


schaze (approved):
Fine by me. I also can't think of a valid use case where the distinction newer or older would be needed. Signaling that it changed is enough.

Issue #323, opened 2025-04-01
Add definitions of device-role and controller-role to the Homie specification


dcj:
As discussed in #322:

> Official descriptions for controllers are documented here: https://homieiot.github.io/implementations/
> 
> I think it would make sense to actually add them to the spec itself. As the spec references device and controller but never clarifies the terms. (It does for a device on a structural level but not necessarily on a role level)

IMO this would be a valuable/important addition to the specification

jacoscaz:
Fully agree!

jacoscaz:
@dcj @schaze this seemed like the perfect entry point for a newcomer so I've opened #325 last week as per https://homieiot.github.io/get_involved/ .

schaze:
@dcj: I merged #325 - please consider closing this issue.
@jacoscaz: Thanks for providing the PR with the addition to the convention!

dcj:
Given nice PR work by @jacoscaz and feedback from @schaze, I'm closing this!

PR #325, opened 2025-04-03
change(definitions): adds top-level 'definitions' section

jacoscaz:
This PR follows up to the discussion in #322 and #323 by adding a top-level section dedicated to definitions at the very top of the spec. The copy itself is taken from both the spec and https://homieiot.github.io/implementations/ .

EDIT 2025-04-07: after speaking with a few other devs interested in Homie, I also added an introductory section for people who might get directed straight to the spec with no prior mentions of what the convention is about.

### DIFF
--- a/convention.md
+++ b/convention.md
@@ -3,6 +3,17 @@
 Version: **<!--VERSION-->x.x.x<!--VERSION-->**
 Date: **<!--DATE-->01. Jan 2000<!--DATE-->**
 
+## Introduction
+
+​The Homie convention is a standardized framework designed to facilitate the
+communication and integration of Internet of Things (IoT) devices using the
+[MQTT protocol](https://mqtt.org).
+
+In particular, the Homie convention defines a consistent topic structure and
+messaging format that enables devices and services to announce themselves and
+their data on an MQTT broker in a uniform manner, supporting automatic device
+discovery.
+
 ## Definitions
 
 - **Device**: an instance of a physical piece of hardware that publishes Nodes and Properties to a MQTT broker. For example, a car, an Arduino/ESP8266 or a coffee machine.
--- a/convention.md
+++ b/convention.md
@@ -10,8 +10,8 @@ communication and integration of Internet of Things (IoT) devices using the
 [MQTT protocol](https://mqtt.org).
 
 In particular, the Homie convention defines a consistent topic structure and
-messaging format that enables devices and services to announce themselves and
-their data on an MQTT broker in a uniform manner, supporting automatic device
+messaging format that enables devices to represent themselves, their data and
+their supported controls/commands in a uniform manner, supporting automatic
 discovery.
 
 ## Definitions
--- a/convention.md
+++ b/convention.md
@@ -147,29 +147,17 @@ A zero length payload published on the `$state` topic indicates a device removal
 
 ## Topology and structure
 
-**Devices:**
-An instance of a physical piece of hardware is called a *device*.
-For example, a car, an Arduino/ESP8266, or a coffee machine.
-Within the convention devices can be modelled to have children. For example, bridge
-devices; a zwave bridge device (the parent) exposes many child devices (the
-zwave devices). There is no depth limit set on additionally nested children.
+The topology of Devices, Nodes and Properties is defined by the following rules:
 
-**Nodes:**
-A *device* can expose multiple *nodes*.
-Nodes are independent or logically separable parts of a device.
-For example, a car might expose a `wheels` node, an `engine` node, and a `lights` node.
+- Within the convention, *devices* can be modelled to have children (**Child Devices**). For example, bridge devices; a zwave bridge device (the parent) exposes many child devices (the zwave devices). There is no depth limit set on additionally nested children.
+- A *device* can expose multiple *nodes*.
+- A *node* can have multiple *properties*.
 
-**Properties:**
-A *node* can have multiple *properties*.
-Properties represent basic characteristics of the node/device, often given as numbers or finite states.
-For example, the `wheels` node might expose an `angle` property.
-The `engine` node might expose a `speed`, `direction`, and `temperature` property.
-The `lights` node might expose an `intensity` and a `color` property.
+### Attributes
 
-**Attributes:**
-*Devices, nodes and properties* have specific *attributes* characterizing them.
-Attributes are represented by a topic identifier starting with `$`.
-The precise definition of attributes is important for the automatic discovery of devices following the Homie convention.
+An **Attribute** is a specific aspect of a *device*, *node* or *property* that is modelled by and directly maps to an MQTT topic. The precise definition of attributes is important for the automatic discovery of devices following the Homie convention.
+
+The convention defines different sets of attributes for *devices*, *nodes* and *properties* respectively. MQTT topic identifiers for topics mapping to attributes always start with `$`.
 
 Examples: A device might have an `IP` attribute, a node will have a `name` attribute, and a property will have a `unit` attribute.
 
--- a/convention.md
+++ b/convention.md
@@ -16,10 +16,15 @@ discovery.
 
 ## Definitions
 
-- **Device**: an instance of a physical piece of hardware that publishes Nodes and Properties to a MQTT broker. For example, a car, an Arduino/ESP8266 or a coffee machine.
-- **Node**: an independent or logically separable part of a device. For example, a car might expose a wheels node, an engine node, and a lights node.
-- **Property**: a basic characteristic of a node. For example, the wheels node might expose an angle property. The engine node might expose a speed, direction, and temperature property. The lights node might expose an intensity and a color property.
-- **Controller**: a software solution that discovers and interacts with Devices through one or more MQTT brokers but does not, by itself, publish MQTT messages.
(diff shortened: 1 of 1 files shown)


dcj:
IMO it would be helpful to distinguish between a "representer of one (or more) fundamental (physical?) entities" (the device-role) and "a software client that discovers, and interacts with those representations" (the controller role)

This distinction is seperate and different from the Device/Note/Property hiearchy.
It is an unfortunate fact-of-life that the term device is overloaded and used to mean two different things, so I propose that when we alway append `-role` to references to devices and controllers.

> **Device**: an instance of a physical piece of hardware that publishes Nodes and Properties to a MQTT broker. For example, a car, an Arduino/ESP8266 or a coffee machine.

The device-role represents one, or more devices to a MQTT broker.  A device-role that is a Matter bridge/gateway/whatever-we-call it, may represent itself as a device, and each proxied Matter thing might be a child-device of the matter-gateway-device.

- **Node**: an independent or logically separable part of a device. For example, a car might expose a wheels node, an engine node, and a lights node.

- **Property**: a basic characteristics of the node/device. For example, the wheels node might expose an angle property. The engine node might expose a speed, direction, and temperature property. The lights node might expose an intensity and a color property.

I propose removing '/device' from node/device above.  A property is representing something about a Node.  And the Node is representing something about a Device....

- **Controller**: a software solution that discovers and interacts with Devices through one or more MQTT brokers but does not, by itself, publish MQTT messages.

A controller can and does publish MQTT messages to settable property topics e.g. `foo/set`
So publishing is not the distinguishing characteristic of a controller-role, it is that the controller is not representing an entity on a MQTT broker.

dcj:
Oops, I was reacting to the first commit, now I see that subsequent commits addressed much, but possibly not all,  of the above.

IDK what git conventions the Homie developers mandate, but in my experience working with other teams, they mostly require squashing commits before pushing/PR....

dcj:
> The Homie convention is a standardized framework

Suggest "open standard framework"

> In particular, the Homie convention defines a consistent topic structure and
> messaging format that enables devices and services to announce themselves and
> their data on an MQTT broker in a uniform manner, supporting automatic device
> discovery.

Suggest "represent" instead of, or in addtion to, "announce"

>  and their data on an MQTT broker in a uniform manner, supporting automatic device
> discovery

their data and supported controls/commands in a uniform manner, supporting automatic discovery and providing self-describing metadata about these respresentations

schaze:
I would keep the nodes and properties definitions where they are. 
On the top only the distinction between controller and device is relevant. Nodes and Props belong to the topology of the device and are not on the same "level" as device and controller.

jacoscaz:
@dcj 

> IDK what git conventions the Homie developers mandate, but in my experience working with other teams, they mostly require squashing commits before pushing/PR....

Personally I prefer cleaning up commits right before _merging_, maintaining the commit history up until then to better understand the thought process behind the proposed changes. That's personal preference, though. I'm happy to go with anything that makes the maintainers happy!

That said, let me take care of the "easy" ones first and ponder about the combined feedback of both you and @schaze 

> Suggest "open standard framework"

Done in https://github.com/homieiot/convention/pull/325/commits/988c1d10dfba747f241acd5efa84db7622344ed2

> Suggest "represent" instead of, or in addtion to, "announce"

> their data and supported controls/commands in a uniform manner, supporting automatic discovery _and providing self-describing metadata about these respresentations_

Done in https://github.com/homieiot/convention/pull/325/commits/87974ed158e5034fb3906a88717335f4835eb1e2 , though I've left out the last part of your suggestion (emphasis mine) as I feel it is overly abstract for an introduction.

> I propose removing '/device' from node/device above. A property is representing something about a Node. And the Node is representing something about a Device....

Done in https://github.com/homieiot/convention/pull/325/commits/8f34b6bac1e9920523b7cea07cff4984a7bde972

jacoscaz:
@schaze 

> I would keep the nodes and properties definitions where they are.
On the top only the distinction between controller and device is relevant.

Done in https://github.com/homieiot/convention/pull/325/commits/d15e5ed5b0800dcafabac22c99eb132c62d31ea5

@dcj 

> This distinction is seperate and different from the Device/Note/Property hiearchy.
> It is an unfortunate fact-of-life that the term device is overloaded and used to mean two different things, so I propose that when we alway append -role to references to devices and controllers.

While I agree with the above, introducing the concept of _roles_ would require more significant changes throughout the spec and that is not something that I feel would be appropriate for this PR, whose goal is merely to find and surface the current definitions of device and controller, however imperfect they might be.

However, in recognition of the overloading of the term "device" that you mention, I thought it prudent to explicitly point it out as part of the definitions (quoting the commit here for posterity just in case we end up making further changes):

https://github.com/homieiot/convention/blob/d15e5ed5b0800dcafabac22c99eb132c62d31ea5/convention.md?plain=1#L17-L27

jacoscaz:
I like how this looks at this very moment. It doesn't try to change things, which was not the point, but (IMHO) it does surface enough information to the top of the document to make for more informed reading. 

Thank you @schaze and @dcj for the wonderful feedback, I hope I have managed to find a good compromise between our sensibilities.

convention.md
```diff
@@ -3,10 +3,35 @@
...
+​The Homie convention is an open standard framework designed to facilitate the
+communication and integration of Internet of Things (IoT) devices using the
+[MQTT protocol](https://mqtt.org).
+
+In particular, the Homie convention defines a consistent topic structure and
+messaging format that enables devices to represent themselves, their data and
+their supported controls/commands in a uniform manner, enabling automatic
+discovery.
+
+## Definitions
```
Tieske: Wouldn't "roles" be a better description? Because there should be way more entries here if we call it definitions.

convention.md
```diff
@@ -3,10 +3,35 @@
...
+
+- **Device**: a physical appliance or logical entity, such as a car, a coffee
+  machine or a protocol bridge, whose representation is published to an MQTT
+  broker.
+- **Controller**: a software solution that discovers and interacts with
+  *devices* through one or more MQTT brokers.
+
+Note that, for brevity and simplicity, the word &quot;device&quot; is often used
+to denote the combination of an appliance, the computer managing the appliance's
+MQTT representation and the representation itself.
```
Tieske, line 30: if you change it to "roles" then this comment can go. And I think we should replace it with a comment that any particular implementation can have either one, or both roles at the same time.

convention.md
```diff
@@ -3,10 +3,35 @@
...
+
 ## MQTT Restrictions
 
 Homie communicates through [MQTT](http://mqtt.org) and is hence based on the basic principles of MQTT topic publication and subscription.
 
+Note that there can be more than one Controller interacting with Devices afferent to the same broker.
+
```
Tieske: Do we need this? this is implicit in the fact that MQTT is the underlying protocol used.

Tieske (changes requested):

convention.md
jacoscaz: Done in https://github.com/homieiot/convention/pull/325/commits/01c93f96ef5ba37f5e42af47e57c45f4ec2ea3ef .

convention.md
jacoscaz: Done in https://github.com/homieiot/convention/pull/325/commits/c1506979acf85b6c45845cbbd74fb7336e4e0c9b .

convention.md
jacoscaz: > And I think we should replace it with a comment that any particular implementation can have either one, or both roles at the same time.

Done in https://github.com/homieiot/convention/commit/c1506979acf85b6c45845cbbd74fb7336e4e0c9b

> if you change it to "roles" then this comment can go

I still feel there are usages of the word "device" that do benefit from this comment. For example, "device ID" isn't (necessarily) an attribute of the _implementation_ playing the _device_ role but, rather, of the _appliance_ being represented.

Generally, I feel that the overloading of the word "device" that @dcj [mentioned](https://github.com/homieiot/convention/pull/325#issuecomment-2783370079) still remains even after changing _definitions_ into _roles_ (which, by the way, is a change I agree with). The goal of this comment is to explicitly embrace this overload to prevent further (and more invasive) changes to the spec (at least for V5).

That said, I'm not married to this comment and I'm happy to remove it if maintainers disagree with the above. I'll leave this conversation up to you @Tieske to resolve once this will have been addressed to your satisfaction.

convention.md
```diff
@@ -3,6 +3,32 @@
...
+## Roles
+
+When interacting through a shared MQTT broker, implementations of the Homie
+convention can take on one or both of the following roles simultaneously:
+
+- **Device**: an implementation that publishes the representation of a physical
+  appliance or logical entity, such as a car, a coffee machine or a protocol
+  bridge, to an MQTT broker.
+- **Controller**: an implementation that discovers and interacts with *devices*
+  through one or more MQTT brokers.
```
```suggestion
- **Controller**: an implementation that discovers and interacts with *devices*
  over MQTT. For example a mobile app, or an if-this-then-that rules engine.
```

Tieske (approved):

Tieske:
lgtm. @schaze please have another look.

schaze:
I'll check it tomorrow - too tired now - and merge it for v5 if everything is fine

schaze (approved):

PR #339, opened 2026-02-11
Add context7 ownership file

Tieske:
see https://context7.com/homieiot/convention

ownership is required to manage versions etc.


schaze (approved):
Good idea!

